PR #173, opened 2023-12-24
fix : unwrap on `Err` inside node_id

### LINKED ISSUE #166: Unwrap causes an error
This occurs when we try to connect to node with wrong node address.

```
2023-12-22T18:46:31.120Z INFO lampod::jsonrpc::peer_control call for `connect` with request `Object {"port": Number(19735), "node_id": String("02b438e989282bb726f575ef356b45163aa37dc074fde9951f122774f34074858"), "addr": String("127.0.0.1")}`. [lampod/src/jsonrpc/peer_control.rs:10]
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidPublicKey', lampo-common/src/model/connect.rs:19:41
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Harshit933:
The application crashes when we try to connect with wrong node id. This commit will fix it.

fix #166

### DIFF
--- a/lampo-common/src/model/connect.rs
+++ b/lampo-common/src/model/connect.rs
@@ -15,8 +15,12 @@ pub struct Connect {
 }
 
 impl Connect {
-    pub fn node_id(&self) -> NodeId {
-        NodeId::from_str(&self.node_id).unwrap()
+    pub fn node_id(&self) -> error::Result<NodeId> {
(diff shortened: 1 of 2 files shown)


vincenzopalazzo:
open again in here https://github.com/vincenzopalazzo/lampo.rs/pull/175

PR #185, opened 2024-02-05
fix : update makefile

Harshit933:
Earlier I noticed that while running `make install` we were getting some error like this :
```
Replaced package `lampod-cli v0.1.0 (/home/harshit/Desktop/lampo.rs/lampod-cli)` with `lampod-cli v0.1.0 (/home/harshit/Desktop/lampo.rs/lampod-cli)` (executable `lampod-cli`)
sudo cp target/debug/liblampo.so /usr/local/lib
[sudo] password for harshit:
cp: cannot stat 'target/debug/liblampo.so': No such file or directory
make: *** [Makefile:22: install] Error 1
```
as we were compiling the lampo in release mode. We were not able to find `target/debug/liblampo.so` as the result was in `target/release/`. Thus, this commit updates the makefile to copy result from `target/release`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ install:
 	$(CC) build --release
 	$(CC) install --locked --path ./lampo-cli 
 	$(CC) install --locked --path ./lampod-cli
-	sudo cp target/debug/liblampo.so /usr/local/lib
+	sudo cp target/release/liblampo.so /usr/local/lib
 
 integration: default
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (approved):
ACK https://github.com/vincenzopalazzo/lampo.rs/pull/185/commits/20b94b865a416e6165fb348ce5dd171bf3773397

PR #188, opened 2024-03-06
fix : add network type inside `getinfo`

### LINKED ISSUE #183: Add network type in getinfo
It would be pretty good to add the network type in the `getinfo` command.

Harshit933:
We simply get the network from `conf` inside the `CoreWalletManager`. Thus, adding a function inside `CoreWalletManager` to return the chain of network the wallet is built upon.

fixes #183

### DIFF
--- a/lampod/src/ln/channel_manager.rs
+++ b/lampod/src/ln/channel_manager.rs
@@ -74,7 +74,6 @@ pub type LampoRouter = DefaultRouter<
 >;
 
 pub struct LampoChannelManager {
-    conf: LampoConf,
     monitor: Option<Arc<LampoChainMonitor>>,
     onchain: Arc<LampoChainManager>,
     wallet_manager: Arc<dyn WalletManager>,
@@ -84,6 +83,7 @@ pub struct LampoChannelManager {
     handler: RefCell<Option<Arc<LampoHandler>>>,
     router: Option<Arc<LampoRouter>>,
 
+    pub(crate) conf: LampoConf,
     pub(crate) channeld: Option<Arc<LampoChannel>>,
     pub(crate) logger: Arc<LampoLogger>,
 }
--- a/lampod/src/ln/inventory_manager.rs
+++ b/lampod/src/ln/inventory_manager.rs
@@ -32,10 +32,12 @@ impl InventoryHandler for LampoInventoryManager {
 
         match event {
             InventoryCommand::GetNodeInfo(chan) => {
+                let chain = self.channel_manager.conf.network.to_string();
                 let getinfo = GetInfo {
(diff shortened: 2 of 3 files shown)


lampod/src/ln/inventory_manager.rs
```diff
@@ -32,10 +32,17 @@ impl InventoryHandler for LampoInventoryManager {
 
         match event {
             InventoryCommand::GetNodeInfo(chan) => {
+                let chain = self
+                    .channel_manager
+                    .onchain()
+                    .wallet_manager
+                    .get_network_type()
+                    .unwrap();
```
vincenzopalazzo, line 40: why did we need to unwrap?

lampod/src/ln/channel_manager.rs
```diff
@@ -194,6 +194,10 @@ impl LampoChannelManager {
         self.channeld.clone().unwrap()
     }
 
+    pub fn onchain(&self) -> Arc<LampoChainManager> {
+        self.onchain.clone()
+    }
```
vincenzopalazzo, line 199: Why?

vincenzopalazzo (changes requested):
Thanks for this, but I think you can access the network directory from the lampo configuration that it is inside the lampod struct?

Or I am missing something?

Harshit933:
> Thanks for this, but I think you can access the network directory from the lampo configuration that it is inside the lampod struct?
> 
> Or I am missing something?

Actually I missed to see conf inside `LampoChannelManager`. Thanks for the review.

lampod/src/ln/inventory_manager.rs
```diff
@@ -32,10 +32,12 @@ impl InventoryHandler for LampoInventoryManager {
 
         match event {
             InventoryCommand::GetNodeInfo(chan) => {
+                let chain = self.channel_manager.conf().network.to_string();
```
vincenzopalazzo, line 35:
```suggestion
                let chain = self.channel_manager.conf.network.to_string();
```

if you can not do that, try `pub (crate) conf`

lampod/src/ln/channel_manager.rs
```diff
@@ -406,6 +406,10 @@ impl LampoChannelManager {
         )));
         Ok(())
     }
+
+    pub fn conf(&self) -> LampoConf {
+        self.conf.clone()
+    }
```
vincenzopalazzo, line 412: imho we do not need it :)

vincenzopalazzo (changes requested):
we can still do better 😄

vincenzopalazzo (approved):
LGMT, thanks

PR #203, opened 2024-04-04
Improve `getinfo`

### LINKED ISSUE #192: Make `getinfo` more awesome
As discussed on discord, we should add more information about the node inside `getinfo`. Make it similar to the `core-lightning` implementation.

Harshit933:
Future fix #192 

- Added `alias`
- Added `blockheight`
- Added `lampo_dir`

### DIFF
--- a/lampo-common/src/conf.rs
+++ b/lampo-common/src/conf.rs
@@ -21,6 +21,7 @@ pub struct LampoConf {
     pub channels_keys: Option<String>,
     pub log_file: Option<String>,
     pub log_level: String,
+    pub alias: Option<String>,
 }
 
 impl LampoConf {
@@ -48,6 +49,7 @@ impl LampoConf {
             channels_keys: None,
             log_level: "info".to_string(),
             log_file: None,
+            alias: None,
         }
     }
 
@@ -213,6 +215,8 @@ impl TryFrom<String> for LampoConf {
             _ => "info".to_string(),
         };
         let log_file = conf.get_conf("log-file").unwrap_or_else(|_| None);
+        let alias = conf.get_conf("alias").unwrap_or(None);
+
         Ok(Self {
             inner: Some(conf),
             root_path,
@@ -227,6 +231,7 @@ impl TryFrom<String> for LampoConf {
             channels_keys,
             log_file,
             log_level: level,
+            alias,
         })
     }
 }
--- a/lampo-common/src/model/getinfo.rs
+++ b/lampo-common/src/model/getinfo.rs
@@ -8,4 +8,5 @@ pub struct GetInfo {
     pub chain: String,
     pub alias: String,
     pub blockheight: u32,
+    pub lampo_dir: String,
 }
--- a/lampo-common/src/model/getinfo.rs
+++ b/lampo-common/src/model/getinfo.rs
@@ -7,5 +7,5 @@ pub struct GetInfo {
     pub channels: usize,
     pub chain: String,
     pub alias: String,
-    pub blockheight: String,
+    pub blockheight: u32,
 }
--- a/lampo-common/src/model/getinfo.rs
+++ b/lampo-common/src/model/getinfo.rs
@@ -6,4 +6,5 @@ pub struct GetInfo {
     pub peers: usize,
     pub channels: usize,
     pub chain: String,
+    pub alias: String,
 }
--- a/lampo-common/src/model/getinfo.rs
+++ b/lampo-common/src/model/getinfo.rs
@@ -7,4 +7,5 @@ pub struct GetInfo {
     pub channels: usize,
     pub chain: String,
     pub alias: String,
+    pub blockheight: String,
 }
--- a/lampod/src/ln/inventory_manager.rs
+++ b/lampod/src/ln/inventory_manager.rs
@@ -41,12 +41,21 @@ impl InventoryHandler for LampoInventoryManager {
                     .map(|alias| alias.to_string());
                 // we have to put "" in case of alias missing as cln provide us with a random alias.
                 let alias = alias.unwrap_or_else(|| "".to_string());
+                let (_, height) = self.channel_manager.onchain.backend.get_best_block()?;
+                let blockheight;
+                if height.is_none() {
+                    blockheight = "Still loading...".to_string();
+                } else {
+                    // Safe to unwrap here
+                    blockheight = height.unwrap().to_string();
+                }
                 let getinfo = GetInfo {
                     node_id: self.channel_manager.manager().get_our_node_id().to_string(),
                     peers: self.peer_manager.manager().list_peers().len(),
                     channels: self.channel_manager.manager().list_channels().len(),
                     chain,
                     alias,
+                    blockheight,
                 };
                 let getinfo = json::to_value(getinfo)?;
                 chan.send(getinfo)?;
--- a/lampod/src/ln/inventory_manager.rs
+++ b/lampod/src/ln/inventory_manager.rs
@@ -38,6 +38,7 @@ impl InventoryHandler for LampoInventoryManager {
                 let alias = alias.unwrap_or_default();
                 let (_, height) = self.channel_manager.onchain.backend.get_best_block()?;
                 let blockheight = height.unwrap_or_default();
+                let lampo_dir = self.channel_manager.conf.root_path.to_string();
                 let getinfo = GetInfo {
                     node_id: self.channel_manager.manager().get_our_node_id().to_string(),
                     peers: self.peer_manager.manager().list_peers().len(),
@@ -45,6 +46,7 @@ impl InventoryHandler for LampoInventoryManager {
                     chain,
                     alias,
                     blockheight,
+                    lampo_dir,
                 };
                 let getinfo = json::to_value(getinfo)?;
                 chan.send(getinfo)?;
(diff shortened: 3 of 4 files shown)


lampo-common/src/conf.rs
```diff
@@ -48,6 +49,7 @@ impl LampoConf {
             channels_keys: None,
             log_level: "info".to_string(),
             log_file: None,
+            alias: "".to_string(),
```
vincenzopalazzo, line 52: nit: there is an optional type in Rust, please use it

lampo-common/src/conf.rs
```diff
@@ -213,6 +215,9 @@ impl TryFrom<String> for LampoConf {
             _ => "info".to_string(),
         };
         let log_file = conf.get_conf("log-file").unwrap_or_else(|_| None);
+        let alias_file = conf.get_conf("alias").unwrap_or(None);
+        let alias = alias_file.unwrap_or_else(|| "".to_string());
```
vincenzopalazzo, line 219: nit: please use optional type

lampo-common/src/conf.rs
```diff
@@ -21,6 +21,7 @@ pub struct LampoConf {
     pub channels_keys: Option<String>,
     pub log_file: Option<String>,
     pub log_level: String,
+    pub alias: String,
```
vincenzopalazzo, line 24: This is too simple, we should produce a better get info

vincenzopalazzo (changes requested):
Mh!

lampo-common/src/conf.rs
Harshit933: I am working on it. This is just one of a step to it.

lampod/src/ln/inventory_manager.rs
```diff
@@ -41,12 +41,21 @@ impl InventoryHandler for LampoInventoryManager {
                     .map(|alias| alias.to_string());
                 // we have to put "" in case of alias missing as cln provide us with a random alias.
                 let alias = alias.unwrap_or_else(|| "".to_string());
+                let (_, height) = self.channel_manager.onchain.backend.get_best_block()?;
+                let blockheight;
+                if height.is_none() {
+                    blockheight = "Still loading...".to_string();
+                } else {
+                    // Safe to unwrap here
+                    blockheight = height.unwrap().to_string();
```
vincenzopalazzo, line 50: no this should be a integer

lampod/src/ln/inventory_manager.rs
```diff
@@ -41,12 +41,21 @@ impl InventoryHandler for LampoInventoryManager {
                     .map(|alias| alias.to_string());
                 // we have to put "" in case of alias missing as cln provide us with a random alias.
                 let alias = alias.unwrap_or_else(|| "".to_string());
+                let (_, height) = self.channel_manager.onchain.backend.get_best_block()?;
+                let blockheight;
+                if height.is_none() {
+                    blockheight = "Still loading...".to_string();
```
vincenzopalazzo, line 47: if it is still loading you should emit a warning

vincenzopalazzo (changes requested):

lampod/src/ln/inventory_manager.rs
Harshit933: Then what should we put up when the blocks are still loading?

vincenzopalazzo: Please see https://github.com/vincenzopalazzo/lampo.rs/pull/203#discussion_r1553820430

In addition, you can check the cln version, the syncing part is in another json field.

P.S: it is not loading, we are not a video app. We are syncing

lampod/src/ln/inventory_manager.rs
Harshit933: But I have added the warning field in ec08e9f7edeee335a041f350117c2f11f3e618fb

vincenzopalazzo: it is wrong :) sorry we can discuss just the feature inside this PR? 

The warnings do not need to be related to the bitcoin core, this is another problem

lampo-nakamoto/src/lib.rs
```diff
@@ -185,4 +185,8 @@ impl Backend for Nakamoto {
     fn process_transactions(&self) -> error::Result<()> {
         unimplemented!()
     }
+
+    fn get_warnings(&self) -> error::Result<String> {
+        unimplemented!()
+    }
```
vincenzopalazzo, line 191:
```suggestion
```

Wrong place, remove all this `get_warnings` call

lampod/src/ln/inventory_manager.rs
```diff
@@ -33,11 +33,32 @@ impl InventoryHandler for LampoInventoryManager {
         match event {
             InventoryCommand::GetNodeInfo(chan) => {
                 let chain = self.channel_manager.conf.network.to_string();
+                let alias = self
+                    .channel_manager
+                    .conf
+                    .alias
+                    .as_ref()
+                    .map(|alias| alias.to_string());
```
vincenzopalazzo, line 41: nit: If you were reading this code instead of writing it, would you think it was insane?

lampod/src/ln/inventory_manager.rs
```diff
@@ -33,11 +33,32 @@ impl InventoryHandler for LampoInventoryManager {
...
+                // we have to put "" in case of alias missing as cln provide us with a random alias.
+                let alias = alias.unwrap_or_else(|| "".to_string());
+                let (_, height) = self.channel_manager.onchain.backend.get_best_block()?;
+                let blockheight;
+                if height.is_none() {
+                    blockheight = 0
+                } else {
+                    // Safe to unwrap here
+                    blockheight = height.unwrap();
+                }
```
vincenzopalazzo, line 51:
```suggestion
                let blockheight = height.unwrap_or_default();
```

Before, you used function chaining excessively, and now, scarcely :)

lampod/src/ln/inventory_manager.rs
```diff
@@ -33,11 +33,32 @@ impl InventoryHandler for LampoInventoryManager {
...
+                let lampo_dir = self.channel_manager.conf.root_path.to_string();
                 let getinfo = GetInfo {
                     node_id: self.channel_manager.manager().get_our_node_id().to_string(),
                     peers: self.peer_manager.manager().list_peers().len(),
                     channels: self.channel_manager.manager().list_channels().len(),
                     chain,
+                    alias,
+                    blockheight,
+                    warnings: self.channel_manager.onchain.backend.get_warnings().unwrap(),
```
vincenzopalazzo, line 60:
```suggestion
```

nit: this is excessive, if there is no good way to add warning, we do not add it. Do not overcomplicate the code

lampo-common/src/model/getinfo.rs
```diff
@@ -6,4 +6,8 @@ pub struct GetInfo {
     pub peers: usize,
     pub channels: usize,
     pub chain: String,
+    pub alias: String,
+    pub blockheight: u32,
+    pub warnings: String,
```
vincenzopalazzo, line 11:
```suggestion
```

vincenzopalazzo (changes requested):
Ok this trivial PR get already too much review.

Fix the last things that we I commented and then mark this ready

lampod/src/ln/inventory_manager.rs
Harshit933: Yes, I also realised that this is wrong. The warnings are from the bitcoin node and should not be put up here. The main aim of this should be like the core-lightning one which only tells the warning in case of syncing the bitcoin.

lampod/src/ln/inventory_manager.rs
Harshit933: Oops.

lampo-common/src/conf.rs
```diff
@@ -213,6 +215,8 @@ impl TryFrom<String> for LampoConf {
             _ => "info".to_string(),
         };
         let log_file = conf.get_conf("log-file").unwrap_or_else(|_| None);
+        let alias = conf.get_conf("alias").unwrap_or(None);
```
vincenzopalazzo, line 218: we need to announce this, but looks like I still need to finish some work here https://github.com/vincenzopalazzo/lampo.rs/pull/202

Lets open a issue and keep track of this

lampo-bitcoind/src/lib.rs
vincenzopalazzo, line 1: there is a way to fixup this commit inside this ba5c8f5c6ba321fcacf5b2cf31163d0ff51f11bd?

I usually use `git commit --fixup ba5c8f5c6ba321fcacf5b2cf31163d0ff51f11bd` before commit but while you have already committed the code.

Maybe if you reset the previous commit `git reset HEAD~` you can still fixup.

try to make some experiments in a new copy of the branch and then we can chat also on discord in the git channel

vincenzopalazzo (changes requested):
LGTM, minors comments only but the code looks good also the git history, really good job

lampo-common/src/conf.rs
vincenzopalazzo: >I don't get it. Why do we need to announce alias? Is this related to bolt12?

we need to announce to the network with a node announcement, it is not related to bolt12, but I did some of the work on the PR that support the bolt12

vincenzopalazzo (approved):
ACK 36c03c2c7899c6c9d0e5fd898f62acc24a8935c8

P.S: remember to open the issue with the alias announcement

PR #204, opened 2024-04-04
fix: crash caused when the port is already in use

### LINKED ISSUE #172: Crash caused when the port is already in use
I was using my lightning node with the same port as my lampo node.

```
2023-12-23T18:52:14.087Z INFO jsonrpc starting server on /home/harshit/.lampo/testnet/lampod.socket. [lampo-jsonrpc/src/lib.rs:178]
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 98, kind: AddrInUse, message: "Address already in use" }', lampod/src/ln/peer_manager.rs:148:22
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

VanshulB:
This pull request fixes #172 , where the application crashes when the specified port is already in use. The crash is occuring cause we are using unwrap() while handling TcpListener::bind().
This commit fixes it by replacing unwrap() with match expressions for error handling.

### DIFF
--- a/lampod/src/ln/peer_manager.rs
+++ b/lampod/src/ln/peer_manager.rs
@@ -148,23 +148,42 @@ impl LampoPeerManager {
         };
         let peer_manager = peer_manager.clone();
         std::thread::spawn(move || {
-            async_run!(async move {
+            let result = async_run!(async move {
                 let bind_addr = format!("0.0.0.0:{}", listen_port);
                 log::info!(target: "lampo", "Litening for in-bound connection on {bind_addr}");
-                let listener = tokio::net::TcpListener::bind(bind_addr).await.unwrap();
+                let listener = match tokio::net::TcpListener::bind(bind_addr).await {
+                    Ok(listener) => listener,
+                    Err(e) => {
+                        return Err::<(), _>(error::anyhow!("Error binding to address: {}", e));
+                    }
+                };
+
                 loop {
                     let peer_manager = peer_manager.clone();
-                    let tcp_stream = listener.accept().await.unwrap().0;
-                    log::info!(target: "lampo", "Got new connection {}", tcp_stream.peer_addr().unwrap());
-                    let _ = tokio::spawn(async move {
-                        // Use LDK's supplied networking battery to facilitate inbound
-                        // connections.
-                        net::setup_inbound(peer_manager.clone(), tcp_stream.into_std().unwrap())
+                    let accept = listener.accept().await;
+                    let accept = accept
+                        .map_err(|err| error::anyhow!("Error accepting connection: {}", err))?;
+                    match accept {
+                        (tcp_stream, _) => {
+                            log::info!(target: "lampo", "Got new connection {}", tcp_stream.peer_addr().unwrap());
+                            let _ = tokio::spawn(async move {
+                                // Use LDK's supplied networking battery to facilitate inbound
+                                // connections.
+                                net::setup_inbound(
+                                    peer_manager.clone(),
+                                    tcp_stream.into_std().unwrap(),
+                                )
+                                .await;
+                            })
                             .await;
-                    })
-                    .await;
+                        }
+                    }
                 }
             });
+            if let Err(err) = &result {
+                log::error!("error while try to listen on inbound connection: `{err}`");
+            }
+            result
         });
         Ok(())
     }


lampod/src/ln/peer_manager.rs
```diff
@@ -141,6 +141,7 @@ impl LampoPeerManager {
         Ok(())
     }
 
+    #[allow(unused_must_use)]
```
vincenzopalazzo, line 144: nit: we should fix the warning and not oppress clippy

lampod/src/ln/peer_manager.rs
```diff
@@ -151,18 +152,36 @@ impl LampoPeerManager {
             async_run!(async move {
                 let bind_addr = format!("0.0.0.0:{}", listen_port);
                 log::info!(target: "lampo", "Litening for in-bound connection on {bind_addr}");
-                let listener = tokio::net::TcpListener::bind(bind_addr).await.unwrap();
+                let listener = match tokio::net::TcpListener::bind(bind_addr).await {
+                    Ok(listener) => listener,
+                    Err(e) => {
+                        log::error!(target: "lampo", "Error binding to address: {}", e);
+                        return Err::<(), error::Error>(e.into());
+                    }
```
vincenzopalazzo, line 160:
```suggestion
                    Err(e) => error::bail!("Error binding to address: {err}")
```

lampod/src/ln/peer_manager.rs
```diff
@@ -151,18 +152,36 @@ impl LampoPeerManager {
...
+                                .await;
+                            })
                             .await;
-                    })
-                    .await;
+                        }
+                        Err(e) => {
+                            log::error!(target: "lampo", "Error accepting connection: {}", e);
+                            return Err::<(), error::Error>(e.into());
+                        }
```
vincenzopalazzo, line 183:
```suggestion
                        Err(e) => error::bail!("Error accepting connection: {err}")
```

lampod/src/ln/peer_manager.rs
```diff
@@ -151,18 +152,36 @@ impl LampoPeerManager {
...
                 loop {
                     let peer_manager = peer_manager.clone();
-                    let tcp_stream = listener.accept().await.unwrap().0;
-                    log::info!(target: "lampo", "Got new connection {}", tcp_stream.peer_addr().unwrap());
-                    let _ = tokio::spawn(async move {
-                        // Use LDK's supplied networking battery to facilitate inbound
-                        // connections.
-                        net::setup_inbound(peer_manager.clone(), tcp_stream.into_std().unwrap())
+
+                    match listener.accept().await {
```
vincenzopalazzo, line 166:
```suggestion
                    match listener.accept().await.map_err(|err| error::anyhow!("Error accepting connection: {err}"))? {
```

I add the previous suggestion by changing your code, but I think you can just use this `map_or` syntax.

In addition, this syntax is also used a lot in the code

vincenzopalazzo (changes requested):
Thanks for the patch and sorry for the delay of the review.

I added some suggestion, but overall I like it 😄

lampod/src/ln/peer_manager.rs
VanshulB: On implementing `error::bail!("Error binding to address: {err}")`the compiler complains about 

```
error[E0282]: type annotations needed
   --> lampod/src/ln/peer_manager.rs:161:17
    |
161 | /                 loop {
162 | |                     let peer_manager = peer_manager.clone();
163 | |
164 | |                     match listener
...   |
182 | |                     }
183 | |                 }
    | |_________________^ cannot infer type

For more information about this error, try `rustc --explain E0282`.
warning: `lampod` (lib) generated 1 warning
```

Could you please tell why is this error is coming, and how can i infer the type of `peer_manager`

lampod/src/ln/peer_manager.rs
```diff
@@ -141,30 +141,32 @@ impl LampoPeerManager {
...
-    #[allow(unused_must_use)]
     pub fn run(&self) -> error::Result<()> {
         let listen_port = self.conf.port;
         let Some(ref peer_manager) = self.peer_manager else {
             error::bail!("peer manager is None, at this point this should be not None");
         };
         let peer_manager = peer_manager.clone();
         std::thread::spawn(move || {
-            async_run!(async move {
+            let result = async_run!(async move {
```
vincenzopalazzo, line 151: nit: this is hiding a huge bug, you never return the error to the caller

```suggestion
            let _ = async_run!(async move {
```

lampod/src/ln/peer_manager.rs
```diff
@@ -177,13 +179,12 @@ impl LampoPeerManager {
...
-                        Err(e) => {
-                            log::error!(target: "lampo", "Error accepting connection: {}", e);
-                            return Err::<(), error::Error>(e.into());
-                        }
                     }
                 }
             });
+            if let Err(err) = result {
+                log::error!(target: "lampo", "{err}");
+            }
```
vincenzopalazzo, line 187: This is a good option to log here but, but we should also return the error to the caller, so use the `?` on the `async_run!`

vincenzopalazzo (changes requested):
Sorry this part of the code is a little bit a mess

lampod/src/ln/peer_manager.rs
VanshulB: Hey, we cannot use `?` to propagate the error to the caller as the compiler says
```
cannot use the `?` operator in a closure that returns `()`
```
so how do i proceed from here, 
should i implement a `mpsc` channel to send errors from the async block back to the main thread

vincenzopalazzo: I need to run the code, I am not sure that the error that you point our is right because you are doing `let res = async_run`, so async_run is not a void.

Most of the time the compiler is wrong

vincenzopalazzo: Please @VanshulB apply

```diff
diff --git a/lampod/src/ln/peer_manager.rs b/lampod/src/ln/peer_manager.rs
index a162e3b..279978a 100644
--- a/lampod/src/ln/peer_manager.rs
+++ b/lampod/src/ln/peer_manager.rs
@@ -161,11 +161,10 @@ impl LampoPeerManager {
                 loop {
                     let peer_manager = peer_manager.clone();

-                    match listener
-                        .accept()
-                        .await
-                        .map_err(|err| error::anyhow!("Error accepting connection: {}", err))?
-                    {
+                    let accept = listener.accept().await;
+                    let accept = accept
+                        .map_err(|err| error::anyhow!("Error accepting connection: {}", err))?;
+                    match accept {
                         (tcp_stream, _) => {
                             log::info!(target: "lampo", "Got new connection {}", tcp_stream.peer_addr().unwrap());
                             let _ = tokio::spawn(async move {
@@ -182,9 +181,10 @@ impl LampoPeerManager {
                     }
                 }
             });
-            if let Err(err) = result {
-                log::error!(target: "lampo", "{err}");
+            if let Err(err) = &result {
+                log::error!("error while try to listen on inbound connection: `{err}`");
             }
+            result
         });
```

vincenzopalazzo:
@VanshulB ok PR looks good except, for the diff that I gave you, it needs to be applied.

In addition, go to the review that I left you, and resolve them if you think that there is nothing to do in this was it is more easy to me understand what you did

vincenzopalazzo:
Looks good and is ready to merge, can I squash the commit in a single one? and fix the commit message in a more structured way?

If for you it is the first time to do this you can find support on the discord 😸

VanshulB:
> Looks good and is ready to merge, can I squash the commit in a single one? and fix the commit message in a more structured way?
> 
> If for you it is the first time to do this you can find support on the discord 😸

Done! 😀

vincenzopalazzo (approved):
ACK https://github.com/vincenzopalazzo/lampo.rs/pull/204/commits/4d4e6506d336c3a404e76e93754c144d1e3d5547

Issue #139, opened 2023-10-29
jsonrpc: deadlock on the server


vincenzopalazzo:
I just fall inside a jsonrpc deadlock server, I think the jsonrpc server needs some more love

vincenzopalazzo:
to prove that this works we should move lnprotest over the UNIX socket too

vincenzopalazzo:
This issue should be reproduced, so we are planning to fix our lnprototest integration testing.

@tareknaser will work on writing a lampo python library, and then I will integrate it inside lnprototest

vincenzopalazzo:
Moving this to the next release in June

vincenzopalazzo:
Moving this to the next release because we have a new version that it is landing to next https://github.com/vincenzopalazzo/lampo.rs/pull/246

Issue #132, opened 2023-10-27
python ffi: crash whire running inside lnprototest


vincenzopalazzo:
We can fix lnprototest to use the Unix socket but I think 
we have a bug inside the python library or inside the lnprototest runner

```
DEBUG    root:lampo.py:102 call to `getinfo` return {'node_id': '03f37129559621cbb4f8f4d4be5dff76ec21c220d7d274a6407683eafb996d97ae', 'peers': 0, 'channels': 0}
INFO     root:runner.py:144 run lampod with node id 03f37129559621cbb4f8f4d4be5dff76ec21c220d7d274a6407683eafb996d97ae
INFO     root:event.py:59 # running {'event': 'Sequence', 'file': 'runner.py', 'pos': '97'}:
DEBUG    root:structure.py:49 receiving event {"event": "Block", "file": "ln_spec_utils.py", "pos": "65"}
INFO     root:event.py:59 # running {'event': 'Block', 'file': 'ln_spec_utils.py', 'pos': '65'}:
DEBUG    root:bitcoind.py:39 Calling getblockcount with arguments ()
DEBUG    root:bitcoind.py:43 Result for getblockcount call: 101
DEBUG    root:bitcoind.py:39 Calling getblockcount with arguments ()
DEBUG    root:bitcoind.py:43 Result for getblockcount call: 101
[...]

PR #205, opened 2024-04-06
Add Lampo Python Client (`pylampo-client`)

tareknaser:
## Description
This pull request implements `pylampo-client`, a Python client for interacting with Lampo through a Unix socket. It uses Poetry for dependency management.

- Defaults to using `<home_dir>/.lampo/regtest/lampod.socket` for the socket path.

**Related issues**: #132, #139
This client should be used in lnprototest integration testing (See https://github.com/vincenzopalazzo/lampo.rs/issues/139#issuecomment-2004988074) 
## Example Usage
```python
from pylampo_client import LampoClient


client = LampoClient()

# Prepare parameters for the "fundchannel" method
params = {
    "node_id": "<node_id>",
    "amount": 100000,
    "public": True,
    "addr": "<address>",
    "port": 19735
}

# Call the "fundchannel" method 
response = client.call("fundchannel", params)
print(response)
```

### DIFF
--- a/tools/pylampo-client/client.py
+++ b/tools/pylampo-client/client.py
@@ -1,0 +1,53 @@
+import json
+import socket
+import os
+
+
+class LampoClient:
+    """
+    A simple Lampo client that communicates via a Unix socket.
+    """
+
+    def __init__(self, socket_path: str = None):
+        """
+        Initializes the LampoClient instance.
+
+        Args:
+          socket_path: (Optional) The path to the Lampo socket.
+                      Defaults to '<home_dir>/.lampo/regtest/lampod.socket'.
+        """
+        if socket_path:
+            self.socket_path = socket_path
+        else:
+            home_dir = os.environ["HOME"]
+            self.socket_path = f"{home_dir}/.lampo/regtest/lampod.socket"
+
+    def call(self, method: str, params: dict = None) -> dict:
+        """
+        Calls a method on the Lampo client over the Unix socket.
+
+        Args:
+          method: The name of the Lampo method to call.
+          params: (Optional) A dictionary of parameters to pass to the method.
+
+        Returns:
+          The response from the Lampo client as a dictionary.
+
+        Raises:
+          Exception: If there is an error communicating with the Lampo client.
+        """
+
+        request = {
+            "method": method,
+            "params": params if params else {},
+            "id": "",
+            "jsonrpc": "2.0",
+        }
+        try:
+            with socket.socket(socket.AF_UNIX, socket.SOCK_STREAM) as sock:
+                sock.connect(self.socket_path)
+                sock.sendall(json.dumps(request).encode())
+                response = sock.recv(1024).decode()
+                return json.loads(response)
+        except Exception as e:
+            raise Exception(f"Error communicating with Lampo client: {e}")
--- a/tools/pylampo-client/client.py
+++ b/tools/pylampo-client/client.py
@@ -1,6 +1,5 @@
 import json
 import socket
-import os
 
 
 class LampoClient:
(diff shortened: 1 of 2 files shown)


tools/pylampo-client/client.py
```diff
@@ -0,0 +1,53 @@
...
+
+        Args:
+          socket_path: (Optional) The path to the Lampo socket.
+                      Defaults to '<home_dir>/.lampo/regtest/lampod.socket'.
+        """
+        if socket_path:
+            self.socket_path = socket_path
+        else:
+            home_dir = os.environ["HOME"]
+            self.socket_path = f"{home_dir}/.lampo/regtest/lampod.socket"
```
vincenzopalazzo, line 23: we can have the `socket_path` required inside the code? this default one do not make really sense because the network is fixed

tools/pylampo-client/client.py
```diff
@@ -0,0 +1,53 @@
...
+          The response from the Lampo client as a dictionary.
+
+        Raises:
+          Exception: If there is an error communicating with the Lampo client.
+        """
+
+        request = {
+            "method": method,
+            "params": params if params else {},
+            "id": "",
```
vincenzopalazzo, line 43:
```suggestion
            "id": "pylampo-client/1",
```

No empty value, but default one

tools/pylampo-client/client.py
```diff
@@ -0,0 +1,53 @@
...
+            "jsonrpc": "2.0",
+        }
+        try:
+            with socket.socket(socket.AF_UNIX, socket.SOCK_STREAM) as sock:
+                sock.connect(self.socket_path)
+                sock.sendall(json.dumps(request).encode())
+                response = sock.recv(1024).decode()
```
vincenzopalazzo, line 50:
```suggestion
                response = ""
                    while True:
                        data = socket.recv(1024)
                        if not data:
                            break
                        response += data.decode() 
```

Your version is buggy and works only for small stream

tools/pylampo-client/pyproject.toml
```diff
@@ -0,0 +1,13 @@
+[tool.poetry]
+name = "pylampo-client"
+version = "0.0.1"
+description = "A simple Lampo client that communicates via a Unix socket"
+authors = ["Tarek <tareknaser360@gmail.com>"]
+
+[tool.poetry.dependencies]
+python = "^3.8"
```
vincenzopalazzo, line 8:
```suggestion
python = "^3.7"
```

I do not see any reason to use 3.8 but not support 3.7. There is any that I am missing?

vincenzopalazzo (changes requested):
Ok I was passing by, and I left a review

tools/pylampo-client/pyproject.toml
tareknaser: Nothing really. this was auto-generated with `poetry`. 
I can make it `3.7`

tools/pylampo-client/client.py
tareknaser: No problem
I thought it may make life easier for testing since we will use it with lnprototest tests

tools/pylampo-client/pyproject.toml
vincenzopalazzo: yeah I was assuming that poetry took your default :)

tools/pylampo-client/client.py
vincenzopalazzo: yes it is bug in test we are pointing to a tmp directory so the home is not really used in the testing one

tools/pylampo-client/client.py
tareknaser: Do you prefer increasing buffer size (4096 maybe?) or receiving data in chunks?

tools/pylampo-client/client.py
tareknaser: Good point

tools/pylampo-client/client.py
tareknaser: How can I decide on a value for buffer size?

vincenzopalazzo: no just use my diff, you can not decide the buffer size you need to read till the end

tareknaser: Thanks!

vincenzopalazzo (approved):
lnprototest uses black for code formatting in python (i am a fan of it) https://github.com/vincenzopalazzo/lampo.rs/blob/main/tests/lnprototest/Makefile#L2 

Maybe we should add it in the future, but for now it is good for me, thanks!

P.S: from now on you are the new mantainer of this package, thinking a way to make it official, but I will redirect to you every problem that there is with this package :)

ACK 67270d69764f417705e51d9be3413abaa326f42d

vincenzopalazzo:
>If we add a CODEOWNERS file later, we can specify that I am responsible for this package

Good idea, can you do it and open a PR? :)

PR #209, opened 2024-04-10
Notify Lampo Daemon of Bitcoin Backend Sync Progress

### LINKED ISSUE #111: bitcoind: notify lampod that the backend is syncing
So, in lightning we have some problems when a node sees some more blocks than us, this makes making impossible
make some operations like funding a channel.

So in the backend we should have an operation that notify in some way the
lampo daemon that the bitcoin backend is syncing and downloading blocks from the network.

This is very common for nodes that are running after a big inactivity period.

VanshulB:
Fixes #111

### DIFF
--- a/lampod/src/chain/blockchain.rs
+++ b/lampod/src/chain/blockchain.rs
@@ -1,10 +1,14 @@
+use std::cell::RefCell;
 use std::collections::HashMap;
 use std::sync::Arc;
 
 use lampo_common::backend::Backend;
-use lampo_common::bitcoin;
+use lampo_common::bitcoin::absolute::Height;
 use lampo_common::bitcoin::blockdata::constants::ChainHash;
 use lampo_common::bitcoin::Transaction;
+use lampo_common::event::onchain::OnChainEvent;
+use lampo_common::event::Event;
+use lampo_common::handler::Handler;
 use lampo_common::ldk;
 use lampo_common::ldk::chain::chaininterface::{
     BroadcasterInterface, ConfirmationTarget, FeeEstimator,
@@ -12,11 +16,17 @@ use lampo_common::ldk::chain::chaininterface::{
 use lampo_common::ldk::chain::Filter;
 use lampo_common::ldk::routing::utxo::UtxoLookup;
 use lampo_common::wallet::WalletManager;
+use lampo_common::{bitcoin, error};
+
+use crate::actions::handler::LampoHandler;
 
 #[derive(Clone)]
 pub struct LampoChainManager {
     pub backend: Arc<dyn Backend>,
     pub wallet_manager: Arc<dyn WalletManager>,
+    pub current_height: RefCell<Height>,
+    pub best_height: RefCell<Height>,
+    pub handler: RefCell<Option<Arc<LampoHandler>>>,
 }
 
 /// Personal Lampo implementation
@@ -27,11 +37,57 @@ impl LampoChainManager {
         LampoChainManager {
             backend: client,
             wallet_manager,
+            handler: RefCell::new(None),
+            // Safe: 0 is a valid consensus height
+            current_height: RefCell::new(Height::from_consensus(0).unwrap()),
+            best_height: RefCell::new(Height::from_consensus(0).unwrap()),
         }
     }
 
     pub fn is_lightway(&self) -> bool {
         self.backend.is_lightway()
+    }
+
+    pub fn set_handler(&self, handler: Arc<LampoHandler>) {
+        self.handler.replace(Some(handler));
+    }
+
+    pub fn handler(&self) -> Arc<LampoHandler> {
+        self.handler.borrow().clone().unwrap()
+    }
+
+    pub fn is_syncing(&self) -> bool {
+        let current_height = self.current_height.borrow();
+        let best_height = self.best_height.borrow();
+        current_height.to_consensus_u32() != best_height.to_consensus_u32()
+    }
+
+    pub fn listen(self: Arc<Self>) -> error::Result<()> {
+        let _ = self.backend.clone().listen()?;
+        let event_listener = self.clone();
+        std::thread::spawn(move || {
+            log::info!(target: "lampo_chain_manager", "listening for chain events");
+            loop {
+                let handler = event_listener.handler();
+                let event = handler.events().recv().unwrap();
+                match event {
+                    Event::OnChain(OnChainEvent::NewBestBlock((_, height))) => {
+                        log::debug!(target: "lampo_chain_manager", "new best block height `{}`", height);
+                        *self.best_height.borrow_mut() = height;
+                    }
+                    Event::OnChain(OnChainEvent::NewBlock(block)) => {
+                        let height = block.bip34_block_height().unwrap();
+                        log::debug!(target: "lampo_chain_manager", "new block with hash `{}` at height `{}`", block.block_hash(), height);
+                        let mut current_height = self.current_height.borrow_mut();
+                        if height as u32 > current_height.to_consensus_u32() {
+                            *current_height = Height::from_consensus(height as u32).unwrap();
+                        }
+                    }
+                    _ => continue,
+                }
+            }
+        });
+        Ok(())
     }
 
     fn print_ldk_target_to_string(&self, target: ConfirmationTarget) -> String {
--- a/lampod/src/ln/inventory_manager.rs
+++ b/lampod/src/ln/inventory_manager.rs
@@ -38,7 +38,9 @@ impl InventoryHandler for LampoInventoryManager {
                 // we have to put "" in case of alias missing as cln provide us with a random alias.
                 let alias = alias.unwrap_or_default();
                 let (_, height) = self.channel_manager.onchain.backend.get_best_block()?;
-                let blockheight = height.unwrap_or_default();
+                let best_height = height.unwrap_or_default();
+                let current_height = self.channel_manager.onchain.current_height.borrow();
+                let block_height = current_height.to_consensus_u32();
                 let lampo_dir = self.channel_manager.conf.root_path.to_string();
                 // We provide a vector here as there may be other types of address in future like tor and ipv6.
                 let mut address_vec = Vec::new();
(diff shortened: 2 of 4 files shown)


lampod/src/ln/channel_manager.rs
```diff
@@ -127,6 +127,15 @@ impl LampoChannelManager {
         self.handler.borrow().clone().unwrap()
     }
 
+    fn is_syncing(&self, block_event_height: u32) -> error::Result<bool> {
+        const SYNC_THRESHOLD: u32 = 5;
+
+        let (_, best_block_height) = self.onchain.backend.get_best_block()?;
```
VanshulB, line 133: Hey @vincenzopalazzo 



Hey @vincenzopalazzo

I would appreciate a bit of guidance here.
cause the method get_best_block is not giving me the latest block height from the network

vincenzopalazzo (commented):
Ok I did not try to implement this solution but what I have in mind could work, and this is the process:

- add inside `LampoChainManager` a `current_height: Arc<Mutex<Height>>`, and also a `best_height`
- inside `LampoChainManager` add a listen method where you call `self.backend.listen()` and you start listening also on a new thread about `OnChainEvent::NewBlock`
- If the best block height is > current height update the value;
- At the same time you listen about `OnChainEvent::NewBestBlock`
- At this point you add inside `LampoChainManager` a method is `is_syncing()` where you compare that `current_height - SYNC_THRESHOLD < best_height`

I sugged that `SYNC_THRESHOLD` will be just 1 or 2

This is an example code that you should use to listen about a incoming events https://github.com/vincenzopalazzo/lampo.rs/blob/main/lampod/src/ln/channel_manager.rs#L135-L166

lampod/src/chain/blockchain.rs
```diff
@@ -27,13 +37,59 @@ impl LampoChainManager {
...
+    pub fn listen(&self) -> error::Result<()> {
+        let _backend_listner = self.backend.clone().listen()?;
+        let event_listener = self.clone();
+        std::thread::spawn(move || {
+            log::info!(target: "lampo_chain_manager", "listening for chain events");
+            loop {
+                let handler = event_listener.handler();
+                let event = handler.events().recv().unwrap();
+                match event {
+                    Event::OnChain(OnChainEvent::NewBlock(block)) => {
```
VanshulB, line 74: Hey @vincenzopalazzo 
Need your help here
I did not see any `NewBlock` event emitted in the logs, only `NewBestBlock` was emitted

lampod/src/chain/blockchain.rs
```diff
@@ -27,13 +37,59 @@ impl LampoChainManager {
         LampoChainManager {
             backend: client,
             wallet_manager,
+            current_height: Arc::new(Mutex::new(Height::from_consensus(0).unwrap())),
+            best_height: Arc::new(Mutex::new(Height::from_consensus(0).unwrap())),
```
vincenzopalazzo, line 41: Please add a safety comment if call unwrap is safe and do not crash the program, otherwise return the error to the caller

lampod/src/chain/blockchain.rs
```diff
@@ -27,13 +37,59 @@ impl LampoChainManager {
...
+    pub fn set_handler(&self, handler: Arc<LampoHandler>) {
+        self.handler.replace(Some(handler));
+    }
+
+    pub fn handler(&self) -> Arc<LampoHandler> {
+        self.handler.borrow().clone().unwrap()
+    }
+
+    pub fn is_syncing(&self) -> bool {
+        const SYNC_THRESHOLD: u32 = 1;
```
vincenzopalazzo, line 59:
```suggestion
        // FIXME: This is a good idea?
        const SYNC_THRESHOLD: u32 = 1;
```

lampod/src/chain/blockchain.rs
```diff
@@ -27,13 +37,59 @@ impl LampoChainManager {
...
+        let current_height = self.current_height.lock().unwrap();
+        let best_height = self.best_height.lock().unwrap();
+        current_height.to_consensus_u32() + SYNC_THRESHOLD < best_height.to_consensus_u32()
+    }
+
+    pub fn listen(&self) -> error::Result<()> {
+        let _backend_listner = self.backend.clone().listen()?;
```
vincenzopalazzo, line 66:
```suggestion
        let _ = self.backend.clone().listen()?;
```

lampod/src/chain/blockchain.rs
vincenzopalazzo: Mh idk we should debug the backend, I am pretty comfortable that I made some mistake there.

lampod/src/chain/blockchain.rs
```diff
@@ -27,13 +37,59 @@ impl LampoChainManager {
...
+                        let height = block.bip34_block_height().unwrap();
+                        log::debug!(target: "lampo_chain_manager", "new block with hash `{}` at height `{}`", block.block_hash(), height);
+                        let mut current_height = event_listener.current_height.lock().unwrap();
+                        if height as u32 > current_height.to_consensus_u32() {
+                            *current_height = Height::from_consensus(height as u32).unwrap();
+                        }
+                    }
+                    Event::OnChain(OnChainEvent::NewBestBlock((_, height))) => {
+                        log::debug!(target: "lampo_chain_manager", "new best block height `{}`", height);
+                        *event_listener.best_height.lock().unwrap() = height;
```
vincenzopalazzo, line 84: thinking here, why we need a `Mutex` here? I do not think that we need and `Mutex` and still note sure there is the need of a pointer (`Arc`)

vincenzopalazzo (changes requested):
This is a nice code, thanks!

I left some comments, let iterate on it a little bit

vincenzopalazzo (commented):
Ok all the commits can be fold inside a single one.

VanshulB:
> Ok all the commits can be fold inside a single one.

done!

vincenzopalazzo (commented):
LGMT, could you add the information inside the getinfo command https://github.com/vincenzopalazzo/lampo.rs/blob/main/lampod/src/jsonrpc/inventory.rs

lampod/src/ln/inventory_manager.rs
```diff
@@ -43,6 +43,7 @@ impl InventoryHandler for LampoInventoryManager {
                     node_id: self.channel_manager.manager().get_our_node_id().to_string(),
                     peers: self.peer_manager.manager().list_peers().len(),
                     channels: self.channel_manager.manager().list_channels().len(),
+                    is_sycning: self.channel_manager.onchain.is_syncing(),
                     chain,
                     alias,
                     blockheight,
```
vincenzopalazzo, line 49: Inside the inventory there if the following code 

```
                let (_, height) = self.channel_manager.onchain.backend.get_best_block()?;
```

but now this solution is not anymore a good solution because we have the `current_height` inside the chain manager

vincenzopalazzo (changes requested):
Still a small change inside the get info and then I will check it out in my local machine :)

lampod/src/ln/inventory_manager.rs
VanshulB: So, rather than creating `is_syncing` method, we can directly compare the 'current_height' with the `blockheight` or we can directly display `current_height` so the user can directly see that the `current_height` is not in sync with the `blockheight`

However, if we create the `is_syncing` method, we can directly call it wherever we first need to check whether the node is still syncing or not.

vincenzopalazzo: I am not talking about the logic you use inside is_syncing, but rather the GetInfo struct.

We should show best_height: blockheight and block_height: current_height.

This way, we inform the user that there is a gap between the best block and the current one.

Additionally, please remove all instances of blockheight and change them to block_height in a separate commit.

I really dislike this single-word convention; we are no longer in the '90s.

lampod/src/ln/inventory_manager.rs
```diff
@@ -56,9 +56,16 @@ impl InventoryHandler for LampoInventoryManager {
...
                     chain,
                     alias,
-                    blockheight,
+                    best_height,
+                    block_height: self
+                        .channel_manager
+                        .onchain
+                        .current_height
+                        .borrow()
+                        .to_consensus_u32(),
```
vincenzopalazzo, line 68: Can you split this call? this make the code hard to read

vincenzopalazzo (changes requested):
Thinking if we should protect with a mutex the RefCell?

VanshulB:
> Thinking if we should protect with a mutex the RefCell?

I don't think so `Mutex` is required in this usecase

Harshit933 (commented):
Okay I observed somethings related to this PR.

1. When the `best_height == block_height` and the node is syncing then the `is_syncing` will return false which shouldn't be the case as our bitcoin node is still syncing. 
```
[doom@archlinux ~]$ lampo-cli --network testnet getinfo
{
  "node_id": "030d7bd5db0d29138bf7d9257deb458354c3151e29e2efb787e81e0ce7c695dca6",
  "peers": 0,
  "channels": 0,
  "chain": "testnet",
  "alias": "doom",
  "block_height": 721150,
  "best_height": 721150,
  "is_sycning": false,
  "lampo_dir": "/home/doom/.lampo",
  "address": [
    {
      "address": "123.12.12.12",
      "port": 39736
    }
  ]
}
```
2. There are some anomalies here. Although, this could be related to the implementation of `lampo-bitcoind`. Like :
```
[doom@archlinux ~]$ lampo-cli --network testnet getinfo
{
  "node_id": "030d7bd5db0d29138bf7d9257deb458354c3151e29e2efb787e81e0ce7c695dca6",
  "peers": 0,
  "channels": 0,
  "chain": "testnet",
  "alias": "doom",
  "block_height": 721150,
  "best_height": 721156,
  "is_sycning": false,
  "lampo_dir": "/home/doom/.lampo",
  "address": [
    {
      "address": "123.12.12.12",
      "port": 39736
    }
  ]
}
```
The `is_syncing` should ideally be true in this case.

lampod/src/ln/inventory_manager.rs
```diff
@@ -38,7 +38,9 @@ impl InventoryHandler for LampoInventoryManager {
                 // we have to put "" in case of alias missing as cln provide us with a random alias.
                 let alias = alias.unwrap_or_default();
                 let (_, height) = self.channel_manager.onchain.backend.get_best_block()?;
-                let blockheight = height.unwrap_or_default();
+                let best_height = height.unwrap_or_default();
```
vincenzopalazzo, line 41: regarding the @Harshit933 review that noted the problem with the `is_syncing` method.

This code is lying, I am not sure why at this point we need a call to `let (_, height) = self.channel_manager.onchain.backend.get_best_block()?;` if we have all the info inside the `self.channel_manager.onchain`. In addition in this case it is not better has `self.onchain`?

vincenzopalazzo:
Closing because inactive

PR #212, opened 2024-04-18
feat : adds `announce-addr`

Harshit933:
This commits adds the ability to announce the IP address where the lampo node can listen for inbound connections.

### DIFF
--- a/lampod/src/ln/peer_manager.rs
+++ b/lampod/src/ln/peer_manager.rs
@@ -132,10 +132,15 @@ impl LampoPeerManager {
             .clone()
             .ok_or(error::anyhow!("channel manager is None"))?;
         let alias = self.conf.alias.clone().unwrap_or_default();
+        let addr = self
+            .conf
+            .announce_addr
+            .clone()
+            .unwrap_or_else(|| "127.0.0.1".to_string());
         std::thread::spawn(move || {
             let result = async_run!(async move {
-                let bind_addr = format!("0.0.0.0:{}", listen_port);
-                log::info!(target: "lampo", "Litening for in-bound connection on {bind_addr}");
+                let bind_addr = format!("{addr}:{listen_port}");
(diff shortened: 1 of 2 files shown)


lampod/src/ln/peer_manager.rs
```diff
@@ -132,10 +132,15 @@ impl LampoPeerManager {
             .clone()
             .ok_or(error::anyhow!("channel manager is None"))?;
         let alias = self.conf.alias.clone().unwrap_or_default();
+        let addr = self
+            .conf
+            .announce_addr
+            .clone()
+            .unwrap_or_else(|| "0.0.0.0".to_string());
```
vincenzopalazzo, line 139:
```suggestion
            .unwrap_or_else(|| "127.0.0.1".to_string());
```

Better do not expose to any IP, this is a privacy leack

lampod/src/ln/peer_manager.rs
```diff
@@ -132,10 +132,15 @@ impl LampoPeerManager {
...
         std::thread::spawn(move || {
             let result = async_run!(async move {
-                let bind_addr = format!("0.0.0.0:{}", listen_port);
-                log::info!(target: "lampo", "Litening for in-bound connection on {bind_addr}");
+                let bind_addr = format!("{}:{}", addr, listen_port);
```
vincenzopalazzo, line 142:
```suggestion
                let bind_addr = format!("{addr}:{listen_port}");
```

This should work too

vincenzopalazzo (changes requested):
Small comments, thanks!

lampod/src/ln/peer_manager.rs
Harshit933: Oops

vincenzopalazzo (approved):
ACK https://github.com/vincenzopalazzo/lampo.rs/pull/212/commits/d6014aec9fb097c7f90dcb9c742ad0817710781c

Thanks!

Issue #211, opened 2024-04-15
ask the seed to the user not as a option parameter


vincenzopalazzo:
Currently we have a privacy leak but we are not on mainet so we are safe.

We should ask the user to tell us the words in a very interactive way and not with command line parameter.

lazyfuhrer:
How would you like the user to tell the words ? @vincenzopalazzo 
example:
```bash
$ lampod-cli --network <network> --restore-wallet
- Enter your BIP-39 seed words one by one.
$ Word (1): book
$ Word (2): pool
....
$ Word (12): nice
```
I'm willing to work on this.

PR #213, opened 2024-04-19
refactor: ask the seed string interactively

lazyfuhrer:
Fixes #211 as discussed in discord

PR #223, opened 2024-05-18
cli: ask as an input the bip 39

vincenzopalazzo:
Before this commit the mnemonic was passed by
command line option and it was a privacy leak because the information was stored inside the terminal history.

With this commit, we ask the seeds as input each time that lampo is run with `--restore-wallet` option.

In the future, this option will be required only the first time of initialization of the wallet.

Fixes: https://github.com/vincenzopalazzo/lampo.rs/issues/211

oversees https://github.com/vincenzopalazzo/lampo.rs/pull/213

### DIFF
--- a/lampod-cli/src/main.rs
+++ b/lampod-cli/src/main.rs
@@ -8,8 +8,7 @@ use std::str::FromStr;
 use std::sync::Arc;
 use std::thread::JoinHandle;
 
-use lampod::jsonrpc::channels::json_close_channel;
-use lampod::jsonrpc::channels::json_list_channels;
+use radicle_term as term;
 
 use lampo_bitcoind::BitcoinCore;
 use lampo_common::backend::Backend;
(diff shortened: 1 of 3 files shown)


lampod-cli/src/main.rs
```diff
@@ -47,7 +46,16 @@ fn main() -> error::Result<()> {
 
 /// Return the root directory.
 fn run(args: LampoCliArgs) -> error::Result<()> {
-    let mnemonic = args.mnemonic.clone();
+    let mnemonic = if args.restore_wallet {
+        let inputs: String = term::input(
+            "BIP 39 Mnemonic",
+            None,
+            Some("To restore the wallet lampo need a BIP39 mnemonic, the words need to be divided by spaces"),
```
Harshit933, line 53:
```suggestion
            Some("To restore the wallet, lampo needs a BIP39 mnemonic with words separated by spaces."),
```

Harshit933 (approved):
ACK 46c4a6f8c52de75204edae6ab555134cd620376f

Just a small comment otherwise everything looks good.

PR #224, opened 2024-05-18
proto: some dry-by fixes that improve the onchain log

vincenzopalazzo:
See commits

### DIFF
--- a/lampo-bitcoind/src/lib.rs
+++ b/lampo-bitcoind/src/lib.rs
@@ -393,7 +393,9 @@ impl Backend for BitcoinCore {
                         tx.txid(),
                     )));
                 }
-                TxResult::Discarded => {}
+                TxResult::Discarded => handler.emit(Event::OnChain(
+                    OnChainEvent::UnconfirmedTransaction(txid.clone()),
+                )),
             }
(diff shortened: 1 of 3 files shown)


lampo-bitcoind/src/lib.rs
```diff
@@ -426,74 +428,72 @@ impl Backend for BitcoinCore {
                 let best_block = self.get_best_block();
                 let Ok((block_hash, height)) = best_block else {
                     // SAFETY: if we are in this block the error will be always not null
-                    log::error!(target: "lampo_bitcoind", "Impossible get the inforamtion of the last besh block: {}", best_block.err().unwrap());
-                    break;
+                    log::error!(target: "bitcoind", "Impossible get the inforamtion of the last besh block: {}", best_block.err().unwrap());
+                    continue;
```
Harshit933, line 432: Ideally we should break out of the loop when we don't get the best block to remove further errors. It is also breaking tests.

vincenzopalazzo: nop because if the bitcoin core it is not ready this function will return an error but we should just be able to retry, no?

PR #225, opened 2024-05-19
tools: export the public API of the package

vincenzopalazzo:
in python there is this concept of module defined by the `__init__.py` where you will go to define the public API.

In this way a user can import the API exposed by the package.


tools/pylampo/pylampo_client/__pycache__/__init__.cpython-312.pyc
tareknaser, line 1: We can remove the cached files from the index since they were already tracked
```sh
git rm -r --cached tools/pylampo/pylampo_client/__pycache__
```

tareknaser (approved):
Looks good to me

tools/pylampo/pylampo_client/__pycache__/__init__.cpython-312.pyc
vincenzopalazzo: Thanks!

vincenzopalazzo:
Push the changes requested by @tareknaser so merging this

PR #229, opened 2024-05-25
feat(getinfo) : specify address inside getinfo

Harshit933:
This commit adds the address inside `getinfo` command. Changelog:
- Adds `NetworkInfo` response inside lampo_common::model
  This is done taking in mind the future work that we do on lampo. 

Sample output for `getinfo` :

```
[doom@archlinux ~]$ lampo-cli --network regtest getinfo
{
  "node_id": "02cc3f4c11351843d755569204a69c48cb4233b788392a13d3cf67bb412b14c90f",
  "peers": 0,
  "channels": 0,
  "chain": "regtest",
  "alias": "doom",
  "blockheight": 105,
  "lampo_dir": "/home/doom/.lampo",
  "address": [
    {
      "address": "127.0.0.1",
      "port": 39736
    }
  ]
}

```

### DIFF
--- a/lampo-common/src/model.rs
+++ b/lampo-common/src/model.rs
@@ -13,7 +13,7 @@ pub use getinfo::GetInfo;
 pub mod request {
     pub use crate::model::close_channel::request::*;
     pub use crate::model::connect::Connect;
-    pub use crate::model::getinfo::GetInfo;
+    pub use crate::model::getinfo::*;
     pub use crate::model::invoice::request::*;
     pub use crate::model::keysend::request::*;
     pub use crate::model::new_addr::request::*;
@@ -25,7 +25,7 @@ pub mod request {
 pub mod response {
     pub use crate::model::close_channel::response::*;
     pub use crate::model::connect::Connect;
-    pub use crate::model::getinfo::GetInfo;
+    pub use crate::model::getinfo::*;
     pub use crate::model::invoice::response::*;
     pub use crate::model::keysend::response::*;
     pub use crate::model::new_addr::response::*;
--- a/lampod/src/ln/inventory_manager.rs
+++ b/lampod/src/ln/inventory_manager.rs
@@ -3,6 +3,7 @@ use std::sync::Arc;
 
 use lampo_common::error;
 use lampo_common::json;
+use lampo_common::model::response::NetworkInfo;
 
 use super::{LampoChannelManager, LampoPeerManager};
 use crate::actions::InventoryHandler;
@@ -39,6 +40,18 @@ impl InventoryHandler for LampoInventoryManager {
                 let (_, height) = self.channel_manager.onchain.backend.get_best_block()?;
                 let blockheight = height.unwrap_or_default();
                 let lampo_dir = self.channel_manager.conf.root_path.to_string();
+                // We provide a vector here as there may be other types of address in future like tor and ipv6.
+                let mut address_vec = Vec::new();
+                let address = self.channel_manager.conf.announce_addr.clone();
+                if let Some(addr) = address {
+                    let port = self.channel_manager.conf.port.clone();
+                    // For now we don't iterate as there is only one type of address.
+                    let address_info = NetworkInfo {
+                        address: addr,
+                        port,
+                    };
+                    address_vec.push(address_info);
+                }
                 let getinfo = GetInfo {
                     node_id: self.channel_manager.manager().get_our_node_id().to_string(),
                     peers: self.peer_manager.manager().list_peers().len(),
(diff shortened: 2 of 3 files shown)


lampod/src/ln/inventory_manager.rs
```diff
@@ -39,6 +40,18 @@ impl InventoryHandler for LampoInventoryManager {
                 let (_, height) = self.channel_manager.onchain.backend.get_best_block()?;
                 let blockheight = height.unwrap_or_default();
                 let lampo_dir = self.channel_manager.conf.root_path.to_string();
+                // We provide a vector here as there may be other types of address in future like tor and ipv6.
+                let address = self
+                    .channel_manager
+                    .conf
+                    .announce_addr
+                    .clone()
+                    .unwrap_or("127.0.0.1".to_string());
```
vincenzopalazzo, line 49:
```suggestion
```

an address can be `None`, do not make any assumption on people address

lampo-common/src/model/network.rs
```diff
@@ -0,0 +1,9 @@
+// This could work as a network command in future like that of `getnetworkinfo` inside bitcoin.
+pub mod response {
+    use serde::{Deserialize, Serialize};
+    #[derive(Debug, Deserialize, Serialize)]
+    pub struct NetworkInfo {
+        pub address: String,
+        pub port: u64,
+    }
+}
```
vincenzopalazzo, line 9: put this code where there is the `GetInfo` code

vincenzopalazzo (changes requested):
Thanks for the reactivity on this problem!

left a couple of review

lampod/src/ln/inventory_manager.rs
Harshit933: If the address is not specified inside the `conf`. We default it to `127.0.0.1` as mentioned inside `peer_menager` :

https://github.com/vincenzopalazzo/lampo.rs/blob/1f92659e1551d8bd94fdea3092f1a0b37ef24524/lampod/src/ln/peer_manager.rs#L139

vincenzopalazzo: @Harshit933 also in this case it is wrong, we should set it to `None` and avoid to bind the server on a specific peer.

However it is unrelated to this PR

Harshit933: Opening this as an new issue

lampo-common/src/model.rs
```diff
@@ -13,7 +13,7 @@ pub use getinfo::GetInfo;
 pub mod request {
     pub use crate::model::close_channel::request::*;
     pub use crate::model::connect::Connect;
-    pub use crate::model::getinfo::GetInfo;
+    pub use crate::model::getinfo::{GetInfo, NetworkInfo};
```
vincenzopalazzo, line 16:
```suggestion
    pub use crate::model::getinfo::*;
```

lampo-common/src/model.rs
```diff
@@ -25,7 +25,7 @@ pub mod request {
 pub mod response {
     pub use crate::model::close_channel::response::*;
     pub use crate::model::connect::Connect;
-    pub use crate::model::getinfo::GetInfo;
+    pub use crate::model::getinfo::{GetInfo, NetworkInfo};
```
vincenzopalazzo, line 28:
```suggestion
    pub use crate::model::getinfo::*;
```

lampod/src/ln/inventory_manager.rs
```diff
@@ -39,6 +40,20 @@ impl InventoryHandler for LampoInventoryManager {
                 let (_, height) = self.channel_manager.onchain.backend.get_best_block()?;
                 let blockheight = height.unwrap_or_default();
                 let lampo_dir = self.channel_manager.conf.root_path.to_string();
+                // We provide a vector here as there may be other types of address in future like tor and ipv6.
+                let mut address_vec = Vec::new();
+                let address = self.channel_manager.conf.announce_addr.clone();
+                if address != None {
+                    // Safe to unwrap
+                    let addr = address.unwrap();
```
vincenzopalazzo, line 48:
```suggestion
                if if let Some(addr) = address {
```

Please use the good feature of rust

vincenzopalazzo (changes requested):
Sometimes you will make trivial mistake that required an additional iteration.

If you read the other code you see a patter that help you avoid this mistakes

lampod/src/ln/inventory_manager.rs
Harshit933: Sorry for that

vincenzopalazzo (approved):
ACK 4fe3965ce16bd21c0cfbe57e96f122d4cbc20a85

PR #234, opened 2024-05-26
Fixes related to `address` inside conf.

### LINKED ISSUE #231: @Harshit933 also in this case it is wrong, we should set it to `None` and avoid to bind the server on a specific peer.
@Harshit933 also in this case it is wrong, we should set it to `None` and avoid to bind the server on a specific peer.

However it is unrelated to this PR

_Originally posted by @vincenzopalazzo in https://github.com/vincenzopalazzo/lampo.rs/pull/229#discussion_r1614519284_

Harshit933:
There are two things that are addressed here : 

1. Throw a warning that the node is not reachable if the `announce-addr` inside `conf` is not present.
2. Stop the node in case the `announce-addr` mentioned inside the `conf` is wrong or could not be binded.
Fixes #231

### DIFF
--- a/lampod-cli/src/main.rs
+++ b/lampod-cli/src/main.rs
@@ -155,7 +155,7 @@ fn run(args: LampoCliArgs) -> error::Result<()> {
         std::process::exit(0);
     })?;
 
-    let workder = lampod.listen().unwrap();
+    let workder = lampod.listen()?;
     log::info!(target: "lampod-cli", "------------ Starting Server ------------");
     let _ = workder.join();
     let _ = jsorpc_worker.join().unwrap();
--- a/lampod/src/ln/peer_manager.rs
+++ b/lampod/src/ln/peer_manager.rs
@@ -132,13 +132,10 @@ impl LampoPeerManager {
             .clone()
             .ok_or(error::anyhow!("channel manager is None"))?;
         let alias = self.conf.alias.clone().unwrap_or_default();
-        let addr = if let Some(addr) = self.conf.announce_addr.clone() {
-            addr
-        } else {
-            log::warn!("Address not specified inside conf, node is not reachable");
+        let Some(addr) = self.conf.announce_addr.clone() else {
             return Ok(());
         };
-        std::thread::spawn(move || {
+        let result = std::thread::spawn(move || {
             let result = async_run!(async move {
                 let bind_addr = format!("{addr}:{listen_port}");
                 log::info!(target: "lampo", "Listening for in-bound connection on {bind_addr}");
@@ -204,6 +201,7 @@ impl LampoPeerManager {
             }
             result
         });
+        result.join().unwrap()?;
         Ok(())
     }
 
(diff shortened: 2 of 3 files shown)


lampod-cli/src/main.rs
```diff
@@ -155,7 +155,9 @@ fn run(args: LampoCliArgs) -> error::Result<()> {
         std::process::exit(0);
     })?;
 
-    let workder = lampod.listen().unwrap();
+    let workder = lampod.listen().map_err(|err| {
+        return error::anyhow!("{err}");
+    })?;
```
vincenzopalazzo, line 160:
```suggestion
    let workder = lampod.listen()?;
```

Please, do not apply monkey change

lampod/src/ln/peer_manager.rs
```diff
@@ -132,12 +132,13 @@ impl LampoPeerManager {
...
-        let addr = self
-            .conf
-            .announce_addr
-            .clone()
-            .unwrap_or_else(|| "127.0.0.1".to_string());
-        std::thread::spawn(move || {
+        let addr = if let Some(addr) = self.conf.announce_addr.clone() {
+            addr
+        } else {
+            log::warn!("Address not specified inside conf, node is not reachable");
```
vincenzopalazzo, line 138:
```suggestion
```

Why a warning? if I want to stay private I do  not announce the address

lampod/src/ln/peer_manager.rs
```diff
@@ -132,12 +132,13 @@ impl LampoPeerManager {
...
+            return Ok(());
+        };
```
vincenzopalazzo, line 140:
```suggestion
        let Some(addr) = self.conf.announce_addr else {
            log::warn!("Address not specified inside conf, node is not reachable");
            return Ok(());
        };
```

Better usage of rust

lampod/src/ln/peer_manager.rs
```diff
@@ -203,6 +204,7 @@ impl LampoPeerManager {
             }
             result
         });
+        result.join().unwrap()?;
```
vincenzopalazzo, line 207: Why this change? you are blocking lampo so from this join the code will never return, Have you tested the code?

vincenzopalazzo (changes requested):

lampod/src/ln/peer_manager.rs
Harshit933: Should I put a `log::info!` or completely removed this line?

vincenzopalazzo: There is no needed to add this message imho

lampod-cli/src/main.rs
Harshit933: Thanks, changed

lampod/src/ln/peer_manager.rs
Harshit933: I still need to see this, Will do in this week!

vincenzopalazzo: Take your time, this will be in the august release

vincenzopalazzo:
Closing because inactive

Issue #232, opened 2024-05-25
create the root directory if it not exist


vincenzopalazzo:
@chrisguida reported this on Signal

```
$ lampod-cli
Error: The path /home/cguida/.lampo doesn't exist
```

`lampod-cli` should be able to create `lampo` dir alone

chrisguida:
I'll do this!

PR #235, opened 2024-05-27
don't error when datadir doesn't exist

chrisguida:
this creates the dir properly downstream
fixes https://github.com/vincenzopalazzo/lampo.rs/issues/232

### DIFF
--- a/lampod-cli/src/args.rs
+++ b/lampod-cli/src/args.rs
@@ -65,11 +65,6 @@ impl TryInto<LampoConf> for LampoCliArgs {
         };
 
         let path = self.data_dir.unwrap_or(conf.root_path);
-        // if the path doesn't exist, return an error
-        if !std::path::Path::new(&path).exists() {
-            error::bail!("The path {} doesn't exist", path);
-        }
-
         // FIXME: this override the full configuration, we should merge the two
         conf = LampoConf::new(Some(path), Some(conf.network), None)?;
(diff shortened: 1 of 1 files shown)


lampod-cli/src/args.rs
```diff
@@ -65,11 +65,6 @@ impl TryInto<LampoConf> for LampoCliArgs {
         };
 
         let path = self.data_dir.unwrap_or(conf.root_path);
-        // if the path doesn't exist, return an error
-        if !std::path::Path::new(&path).exists() {
-            error::bail!("The path {} doesn't exist", path);
-        }
```
vincenzopalazzo, line 71: Mh I am wondering if we should check if the directory exist, and if not we will create it?

I am sure that for the network directory inside the `.lampo` we do the same

chrisguida: Have you tried running this branch? That's what it's doing :)

chrisguida: Here's the line where it happens https://github.com/vincenzopalazzo/lampo.rs/blob/main/lampo-common/src/conf.rs#L70

vincenzopalazzo (approved):
Good catch!

Thanks for the contribution :)

ACK https://github.com/vincenzopalazzo/lampo.rs/pull/235/commits/df39ecf8c698ab418c9e2a53a4992d8fc4f7117c

PR #239, opened 2024-05-29
Clean up README

chrisguida:
Clean up README and add warning about Lampo forgetting the seed on restart.
Also fix some typos in log messages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 
 ## Crates
 
-These are the complete list of crates supported right now:
+This is the complete list of crates supported right now:
 
 | Crate       | Description                                   | Version     |
 |:------------|:---------------------------------------------:|------------:|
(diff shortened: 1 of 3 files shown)


vincenzopalazzo (approved):
Thanks!

ACK https://github.com/vincenzopalazzo/lampo.rs/pull/239/commits/2cf4448798db8c181624cbab807eb024d31d32d7

Next time, could you add the changes in separate commits? For example:

1. Change the README
2. Fix the log typos

This way, if we want to create an LTS version like the kernel, we can cherry-pick the commits easily into another branch without causing a mess of changes. 😸

However, thanks for your typo fixes!

chrisguida:
Sure! The reason I did it this way is because the typos in the log messages were referenced in the README, it seemed weird to fix them in the README, then fix them in the place the README is referencing in different commits.

vincenzopalazzo:
Ah make sense! but it is better keep separate also if related, in this way we can cherry-pick just the code that we need

chrisguida:
Ok I'll keep that in mind for future contributions, thanks for providing feedback :)

Issue #199, opened 2024-03-22
Support '?' for RPC errors.


Harshit933:
For now each time `.map_err(|err| {})` looks odd.

vincenzopalazzo:
I should improve this shit, for this release!

Assign to my self

PR #259, opened 2024-06-19
core: improve error handling with anyhow

vincenzopalazzo:
This is a fix step on improving the error handling, but we still need to find out how to improve the error handling for external error type. Probably a good solution for this is thierror.

Suggested-by: @Harshit933
Fixes https://github.com/vincenzopalazzo/lampo.rs/issues/199


lampo-common/src/conf.rs
```diff
@@ -97,10 +98,10 @@ impl LampoConf {
                 // so it is safe unwrap here otherwise we are hiding a bug
                 // and we must crash.
                 .or_else(|| root_path.strip_suffix(&suffix_without_slash))
-                .expect(&format!("path: {root_path} - network: {network}"))
+                .unwrap_or_else(|| panic!("path: {root_path} - network: {network}"))
```
Harshit933, line 101: Why are we panicking here?

Harshit933 (approved):
Single comment only otherwise LGTM

lampo-common/src/conf.rs
vincenzopalazzo: Good question, because this should never fails

PR #277, opened 2024-07-18
feat(lsp): Add support for `lsp` in lampo

Harshit933:
This project aims to configure lampo to be able to act as a lightning liquidity provider and be able to act both as a client and as a server.

Changes done:
1. `lampo-common/src/conf.rs`: Configure lampo config to be able to act as provider or client based on the input we give.
2. `lampo-common/src/event/liquidity.rs`: Created a new event file where we handle all the incoming events from the `lightning-liquidity`
3. `lampod/src/actions/handler.rs`: Made changes to emit different events while we act as a server, events could be: `HTLCIntercepted`, `PaymentForwarded`, `HTLCHandlingFailed` etc.
4. `lampod/src/lib.rs`: Made suitable changes to spin up our node with different configs for `message_handler`.
5. `lampod/src/ln/message_handler.rs`: This was mostly copied from the https://github.com/tnull/ldk-node-hack as it serves two different handlers for peer manager, when we are not acting as a LSP we handle the messages via `LampoCustomMessageHandler::Liquidity` and `LampoCustomMessageHandler::Ignoring` is basically a dummy handler.
6. `tests/tests/src/lampo_tests.rs`: Created two new tests for the LSPs.


lampod/src/liquidity.rs
vincenzopalazzo, line 1: this file need to be inside `lampod/src/ln` and called `liquidity_manager.rs`

lampod/src/liquidity.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use lampo_common::keys::LampoKeysManager;
+use lightning_liquidity::{events::Event, LiquidityManager};
+
+use crate::{chain::LampoChainManager, ln::LampoChannel};
+
+type LampoLiquidity =
+    LiquidityManager<Arc<LampoKeysManager>, Arc<LampoChannel>, Arc<LampoChainManager>>;
+
+struct LampoLiquiditySource {
```
vincenzopalazzo: `LampoLiquiditySource` is a bad name, and it is clear that you copy it somewhere like https://github.com/tnull/ldk-node-hack/blob/fbad621a89b298d9bc7e8a56badddb2094284d2a/src/liquidity.rs#L460 but in Lampo this is `LampoLiquidityManager`

Otherwise, you should justify why and what does mean `Source`

lampod/src/liquidity.rs
```diff
@@ -0,0 +1,35 @@
...
+    lampo_liquidity_manager: Arc<LampoLiquidity>,
+}
+
+impl LampoLiquiditySource {
+    fn new(liquidity: Arc<LampoLiquidity>) -> Self {
+        LampoLiquiditySource {
+            lampo_liquidity_manager: liquidity,
+        }
```
vincenzopalazzo:
```
        Self {
            lampo_liquidity_manager: liquidity,
        }
```

Better this syntax

lampod/src/liquidity.rs
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    pub fn clean_events(&self) {
+        self.lampo_liquidity_manager.get_and_clear_pending_events();
+    }
```
vincenzopalazzo: are you sure that this function does not need the return type?

lampod/src/lib.rs
```diff
@@ -106,6 +113,45 @@ impl LampoDaemon {
         &self.conf
     }
 
+    pub fn liquidity(&self) -> Arc<LampoLiquidityManager> {
+        self.liquidity.clone().unwrap()
+    }
+
+    // This should not be initiated when we open our node only when we
+    // provide some args inside cli or in lampo.conf
+    fn init_lampo_liquidity(
```
vincenzopalazzo, line 121: follow the name convention of other methods

lampod/src/lib.rs
```diff
@@ -106,6 +113,45 @@ impl LampoDaemon {
...
+        &self,
+        node_id: PublicKey,
+        addr: SocketAddress,
+        token: Option<String>,
+    ) -> error::Result<LampoLiquidityManager> {
+        log::info!("Starting lampo-liquidity manager as a consumer!");
+        //TODO: Fix this!
```
vincenzopalazzo, line 128: express the reason for `Fix this`, what you should fix?

lampod/src/lib.rs
```diff
@@ -106,6 +113,45 @@ impl LampoDaemon {
...
+            self.offchain_manager().key_manager(),
+            self.channel_manager().manager(),
+            Some(self.onchain_manager()),
+            None,
+            None,
+            None,
+        );
+        let liquidity = LampoLiquidityManager::new(
+            liquidity.into(),
+            Arc::new(self.conf.clone()),
```
vincenzopalazzo, line 139: `Arc::new(self.conf.clone())` does not look right, why do you need to pass a pointer?

lampod/src/lib.rs
```diff
@@ -106,6 +113,45 @@ impl LampoDaemon {
...
+            Some(LSPProvider {
+                node_id,
+                addr,
+                token,
+                opening_params: None,
+                ctlv_exiry: None,
+                scid: None,
+            }),
+            self.channel_manager.as_ref().unwrap().clone(),
```
vincenzopalazzo, line 148: `self.channel_manager.as_ref().unwrap().clone()` looks wrong, you should have a get method like `self.offchain_manager()`

lampod/src/liquidity.rs
```diff
@@ -1,35 +1,271 @@
-use std::sync::Arc;
+use std::{sync::Arc, time::Duration};
```
vincenzopalazzo, line 1: The import is wrong, you are not following the standard of lampo

lampod/src/liquidity.rs
```diff
@@ -1,35 +1,271 @@
...
+        sign::KeysManager,
+    },
+    secp256k1::{PublicKey, Secp256k1},
+};
+use lightning_liquidity::{
+    events::Event,
+    lsps0::ser::RequestId,
+    lsps2::{event::LSPS2ClientEvent, msgs::OpeningFeeParams},
+    LiquidityManager,
+};
```
vincenzopalazzo: Same here the import are bad

lampod/src/liquidity.rs
```diff
@@ -1,35 +1,271 @@
...
 
-type LampoLiquidity =
+pub type LampoLiquidity =
     LiquidityManager<Arc<LampoKeysManager>, Arc<LampoChannel>, Arc<LampoChainManager>>;
 
-struct LampoLiquiditySource {
-    lampo_liquidity_manager: Arc<LampoLiquidity>,
+// Configured when we are acting as a client sourcing lsp from
+// a different node, here provider is someone who is providing us liquidity.
+pub struct LSPProvider {
```
vincenzopalazzo, line 32: I do not need to read the comment to see that the `LSPProvider` is `LSPClient`, the `P` in the `LSP` is already mean `Provider`

lampod/src/liquidity.rs
```diff
@@ -1,35 +1,271 @@
...
+    pub addr: SocketAddress,
+    pub node_id: PublicKey,
+    pub token: Option<String>,
+    // This will be initialised when we get the response of get_info request
+    pub opening_params: Option<Vec<OpeningFeeParams>>,
```
vincenzopalazzo, line 37: not sure that the comment here is super helpful I still not get why we need it

lampod/src/liquidity.rs
```diff
@@ -0,0 +1,283 @@
...
+                }
+
+                // TODO: Handle this in a better way as we can get new opening_params from a
+                // LSP if it fails to responds within a certain time
+                if self.lsp_provider.as_ref().unwrap().opening_params.is_some() {
+                    log::info!("We already have some params inside lsp_provider")
+                }
+
+                self.lsp_provider.as_mut().unwrap().opening_params = Some(opening_fee_params_menu)
+            }
```
Harshit933, line 124: @vincenzopalazzo For example this event occurs when we call the function `client_request_opening_params`, this contains the information about the a LSP fee rates. So, how do we wait for this event to happen and then proceed with the program flow.

lampod/src/liquidity.rs
```diff
@@ -169,28 +166,48 @@ impl LampoLiquidityManager {
...
+        // TODO: We need to provide a suitable algorithm to get the best_params from all the
+        // opening params that we get from the peer. For now we are getting the first param
+        let best_fee_param = self
+            .lsp_provider
+            .as_ref()
+            .unwrap()
+            .opening_params
+            .as_ref()
+            .unwrap()[0]
+            .clone();
```
vincenzopalazzo, line 195: when you will stop chaining 300 method? This is bad for readiblity

vincenzopalazzo: plus `.opening_params
            .as_ref()
            .unwrap()[0]` looks wrong

lampod/src/liquidity.rs
vincenzopalazzo: still your question is unclear, are you looking some code example that should be included inside your question?

vincenzopalazzo: Idk I should look inside the code

Harshit933: Okay I got a work around

lampod/src/liquidity.rs
Harshit933: This was already updated inside ba24141d32415272030e1c391e680c514f9a2ece

lampod/src/liquidity.rs
Harshit933: Nice catch, thanks. Something like `LampoLiquidityProvider` would work?

lampod/src/lib.rs
Harshit933: Thanks

lampod/src/liquidity.rs
Harshit933: Thanks, changing

lampod/src/lib.rs
Harshit933: Thanks

lampod/src/liquidity.rs
Harshit933: Removed

lampod/src/lib.rs
Harshit933: Removed

lampod/src/liquidity.rs
vincenzopalazzo: The problem is the suffix `Provider`, iirc we already discuss this to call `LSPManager`

lampod/src/liquidity.rs
vincenzopalazzo: Well I would be happy to have a decent comment that explain what is this opening params and why we need it there

lampo-common/src/conf.rs
```diff
@@ -120,6 +127,11 @@ impl LampoConf {
         let input_path = path;
         let path = Self::normalize_root_dir(&conf.root_path, conf.network);
         conf.root_path = path.clone();
+        // Must be used when we act as a liquidity provider
+        conf.ldk_conf = UserConfig {
+            accept_intercept_htlcs: true,
+            ..Default::default()
+        };
```
vincenzopalazzo, line 134: I do not think this is the correct place

lampo-common/src/conf.rs
```diff
@@ -133,6 +145,16 @@ impl LampoConf {
 
         Ok(conf)
     }
+
+    // These functions should be called when we get something like
+    // liquidity=consumer of liquidity=provider inside lampo.conf
+    pub fn configure_as_liquidity_consumer(&mut self) {
+        self.liquidity = Some("Consumer".to_string())
```
vincenzopalazzo, line 152:
```suggestion
        self.liquidity = Some("Consumer".to_string())
```
use an enum like `LiquisityProvider::Consumer`

lampo-common/src/event.rs
```diff
@@ -7,6 +8,7 @@ use std::sync::{Arc, Mutex};
 use crate::chan;
 use crate::event::ln::LightningEvent;
 use crate::event::onchain::OnChainEvent;
+use liquidity::LiquidityEvent;
```
vincenzopalazzo, line 11: wrong place for the import

lampo-common/src/event/liquidity.rs
```diff
@@ -0,0 +1,56 @@
+use bitcoin::secp256k1::PublicKey;
+use lightning::events::HTLCDestination;
+use lightning::ln::{channelmanager::InterceptId, ChannelId, PaymentHash};
+use lightning_liquidity::{lsps0::ser::RequestId, lsps2::msgs::OpeningFeeParams};
+
+#[derive(Debug, Clone)]
+pub enum LiquidityEvent {
```
vincenzopalazzo, line 7: not sure that this is the correct place where to put the event, but maybe this can be moved later in another crate

lampo-common/src/lib.rs
```diff
@@ -48,3 +48,7 @@ pub mod btc_rpc {
         pub mempoolminfee: f32,
     }
 }
+
+pub mod chrono {
```
vincenzopalazzo, line 52: not understand why we need this

lampo-common/src/logger.rs
```diff
@@ -8,7 +8,7 @@ use std::time::SystemTime;
 // FIXME: this is not async we should modify it
 use std::fs::File;
 
-use chrono::prelude::*;
+use crate::chrono::prelude::*;
```
vincenzopalazzo, line 11:
```suggestion
use chrono::prelude::*;
```

lampo-testing/src/lib.rs
```diff
@@ -62,11 +65,13 @@ macro_rules! wait {
 pub struct LampoTesting {
     inner: Arc<LampoHandler>,
     root_path: Arc<TempDir>,
+    liquidity: Option<Arc<LampoLiquidityManager>>,
```
vincenzopalazzo, line 68: not sure that this is the right place

lampo-testing/src/lib.rs
```diff
@@ -62,11 +65,13 @@ macro_rules! wait {
...
     pub port: u64,
     pub wallet: Arc<dyn WalletManager>,
     pub mnemonic: String,
     pub btc: Arc<BtcNode>,
     pub info: response::GetInfo,
+    pub lampod: Option<Arc<LampoDaemon>>,
```
vincenzopalazzo, line 74: we already have the handler, `LampoHandler` should be enough?

lampo-testing/src/lib.rs
```diff
@@ -134,6 +139,103 @@ impl LampoTesting {
...
+            root_path: dir.into(),
+            info,
+            liquidity: None,
+            lampod: None,
+        })
+    }
+
+    // Use this for liquidity.
+    // FIXME: Integrate this with new method
+    pub fn new_liquidity(
```
vincenzopalazzo, line 157: too much code duplication, what you are trying to do?

lampo-testing/src/lib.rs
```diff
@@ -176,4 +280,8 @@ impl LampoTesting {
     pub fn root_path(&self) -> Arc<TempDir> {
         self.root_path.clone()
     }
+
+    pub fn liquidity(&self) -> Arc<LampoLiquidityManager> {
```
vincenzopalazzo, line 284: this should be not needed if you have the handler already

lampod/src/actions/handler.rs
```diff
@@ -31,7 +34,7 @@ pub struct LampoHandler {
     wallet_manager: Arc<dyn WalletManager>,
     chain_manager: Arc<LampoChainManager>,
     external_handlers: RefCell<Vec<Arc<dyn ExternalHandler>>>,
-    #[allow(dead_code)]
+    liquidity_manager: Option<Arc<LampoLiquidityManager>>,
```
vincenzopalazzo, line 37: ok you already have it inside the handler, so in lampo testing you do not need it

lampod/src/lib.rs
```diff
@@ -67,7 +76,7 @@ pub struct LampoDaemon {
     persister: Arc<LampoPersistence>,
     handler: Option<Arc<LampoHandler>>,
     process: Cell<Option<BackgroundProcessor>>,
-
+    liquidity: Option<Arc<LampoLiquidityManager>>,
```
vincenzopalazzo, line 79: move is under the offchain_manager

lampod/src/lib.rs
```diff
@@ -167,16 +181,88 @@ impl LampoDaemon {
...
         log::debug!(target: "lampo", "init peer manager ...");
-        let mut peer_manager = LampoPeerManager::new(&self.conf, self.logger.clone());
+        let mut peer_manager;
+        let peer_liquidity;
+
+        if let Some(liq) = &self.liquidity {
+            peer_liquidity = Some(Arc::clone(&self.liquidity.clone().unwrap()));
+        } else {
+            peer_liquidity = None;
+        }
```
vincenzopalazzo, line 191:
```suggestion
        let peer_manager = if let Some(liq) = &self.liquidity {
            peer_liquidity = Some(Arc::clone(&self.liquidity.clone().unwrap()));
        } else {
            peer_liquidity = None;
        }
```

use this syntax, the one that you are using is really unusual to rust

lampod/src/lib.rs
```diff
@@ -219,11 +305,15 @@ impl LampoDaemon {
         self.init_onchaind(client.clone())?;
         self.init_channeld()?;
         self.init_offchain_manager()?;
+        self.init_liquidity_manager()?;
```
vincenzopalazzo, line 308: explain why this is the right position?

lampod/src/ln/channel_manager.rs
```diff
@@ -58,7 +58,8 @@ pub type LampoArcChannelManager<M, T, F, L> = ChannelManager<
     Arc<L>,
 >;
 
-type LampoChannel =
+// Required inside LampoLiquidity
```
vincenzopalazzo, line 61: the comment does not explain why and when

lampod/src/ln/message_handler.rs
```diff
@@ -0,0 +1,94 @@
...
+use std::sync::Arc;
+
+use lampo_common::bitcoin::secp256k1::PublicKey;
+use lampo_common::ldk::ln::features::{InitFeatures, NodeFeatures};
+use lampo_common::ldk::ln::peer_handler::CustomMessageHandler;
+use lampo_common::ldk::ln::wire::CustomMessageReader;
+
+use lightning_liquidity::lsps0::ser::RawLSPSMessage;
+
+pub enum LampoCustomMessageHandler {
```
vincenzopalazzo, line 12: good implementation

lampod/src/ln/offchain_manager.rs
```diff
@@ -59,6 +59,10 @@ impl OffchainManager {
         })
     }
 
+    pub fn key_manager(&self) -> Arc<LampoKeysManager> {
+        self.keys_manager.clone()
+    }
```
vincenzopalazzo, line 64: should be already exposed by the onchain manager?

lampod/src/ln/peer_manager.rs
```diff
@@ -44,27 +46,33 @@ pub type SimpleArcPeerManager<M, T, L> = PeerManager<
     Arc<P2PGossipSync<Arc<NetworkGraph<Arc<L>>>, Arc<T>, Arc<L>>>,
     Arc<LampoArcOnionMessenger<L>>,
     Arc<L>,
-    IgnoringMessageHandler,
+    LampoCustomMessageHandler,
```
vincenzopalazzo, line 49:
```suggestion
    LampoMessageHandler,
```

Renaming

lampod/src/ln/peer_manager.rs
```diff
@@ -44,27 +46,33 @@ pub type SimpleArcPeerManager<M, T, L> = PeerManager<
...
     Arc<LampoKeysManager>,
 >;
 
-type InnerLampoPeerManager =
+pub type InnerLampoPeerManager =
```
vincenzopalazzo, line 53: Why?

tests/tests/src/lampo_tests.rs
```diff
@@ -509,3 +518,235 @@ pub fn decode_offer() -> error::Result<()> {
...
+            }
+
+            if channels.channels.first().unwrap().ready {
+                return Ok(());
+            }
+        }
+        node3.fund_wallet(6).unwrap();
+        Err(())
+    });
+    // This would return a timeout as json_pay will continue to wait till it get an
```
vincenzopalazzo, line 734: where is the code where you are calling `accept_inbound_channel_from_trusted_peer_0conf`?

lampod/src/actions/handler.rs
```diff
@@ -131,14 +135,23 @@ impl Handler for LampoHandler {
                 push_msat,
                 channel_type,
             } => {
-                Err(error::anyhow!("Request for open a channel received, unfortunatly we do not support this feature yet."))
+                log::info!("Recieved open channel request!");
+                let liquidity_manager = self.liquidity_manager.clone();
+                if let Some(_) = liquidity_manager {
+                    let _ = self.channel_manager.channeld().accept_inbound_channel_from_trusted_peer_0conf(&temporary_channel_id, &counterparty_node_id, 0);
+                }
```
vincenzopalazzo, line 142: I do not think this is called at any point. We should keep the error here

lampod/src/actions/handler.rs
```diff
@@ -131,14 +135,23 @@ impl Handler for LampoHandler {
...
             }
             ldk::events::Event::ChannelReady {
                 channel_id,
                 user_channel_id,
                 counterparty_node_id,
                 channel_type,
             } => {
+                let liquidity_manager = self.liquidity_manager.clone();
+                if let Some(ref liq_manager) = liquidity_manager {
+                    let _ = liquidity_manager.unwrap().channel_ready(user_channel_id, &channel_id, &counterparty_node_id);
```
vincenzopalazzo, line 153: we should do this only with the LSP not with any nodes

vincenzopalazzo (changes requested):

lampod/src/actions/handler.rs
Harshit933: Yes, we are doing this when we have some liquidity config inside the `lampo.conf`

lampod/src/actions/handler.rs
Harshit933: It is being called when we specifically accepting the channels with `0conf`, which would be used when we are a client requesting for in bound liquidity. This would come handy when the payer pays the JIT invoice and the LSP on the go creates a channel with the client.

vincenzopalazzo:
This need to be reworked a little bit !

Issue #256, opened 2024-06-18
ci: try to reproduce the gossip issue reported in ldk


vincenzopalazzo:
We have all the info to reproduce this https://github.com/lightningdevkit/rust-lightning/issues/3075

Maybe we do it after release

PR #279, opened 2024-07-22
jsonrpc: add the possibility to fetch the network graph info

vincenzopalazzo:
This is going to reproduce https://github.com/vincenzopalazzo/lampo.rs/issues/256

Fixes https://github.com/vincenzopalazzo/lampo.rs/issues/256

To run the test consider running `nix develop --command bash -c 'make integration ARGS="-- --test-threads 1 test_sync_gossip_from_network_cln" TEST_LOG_LEVEL=info'` without wasting time in building the test environment.

### DIFF
--- a/lampod/src/jsonrpc/inventory.rs
+++ b/lampod/src/jsonrpc/inventory.rs
@@ -1,5 +1,6 @@
 //! Inventory method implementation
 use lampo_common::json;
+use lampo_common::model::response::{NetworkChannel, NetworkChannels};
 use lampo_jsonrpc::errors::Error;
 
 use crate::LampoDaemon;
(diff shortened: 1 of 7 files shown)


lampod/src/jsonrpc/inventory.rs
```diff
@@ -9,3 +10,20 @@ pub fn get_info(ctx: &LampoDaemon, request: &json::Value) -> Result<json::Value,
...
+pub fn json_network_channels(ctx: &LampoDaemon, _: &json::Value) -> Result<json::Value, Error> {
+    let network_graph = ctx.channel_manager().graph();
+    let network_graph = network_graph.read_only();
+    let channels = network_graph.channels().unordered_keys();
+    let mut network_channels = Vec::new();
+    for short_id in channels {
+        let Some(channel) = network_graph.channel(*short_id) else {
+            continue;
+        };
+        network_channels.push(NetworkChannel::from(channel.clone()));
```
vincenzopalazzo, line 24: we should avoid to clone here

PR #294, opened 2024-10-17
ldk-node: introduce the ldk node as a external handler

vincenzopalazzo:
Just playing a little bit on the flexibility of the lampo design


vincenzopalazzo:
Making ldk-node an external handler is fine but will couple the ldk-node with lampo, and this make a mess with how we init the node regarding the ldk and lampo.

Probably will be just good to have the lampod-cli be able to start a lampo-ldk-node. We should have just a way to make some conversion for the ldk-node calls

PR #284, opened 2024-08-27
Adding Support for VLS Signer

VanshulB:
This pull request removes the LDK InMemorySigner and replaces it with VLS Signer
Lampo will listen to `vlsd2` on port 6600 by default
you can change this port by passing `vls-port` argument as well

```
> lampod-cli --network regtest
```
and then run the vlsd2
```
>  ./vlsd2 --network=regtest --rpc-user <core-user> --rpc-pass <core-password>  --policy-filter=policy-commitment-htlc-routing-balance:warn --connect  http://127.0.0.1:6600
```

PR #295, opened 2024-10-30
--restore-wallet restores words from file

Psycho-Pirate:
Added functionality to store words everytime new wallet is generated or wallet is restored by user.

--restore-wallet reads from the file if it exists, otherwise asks for words from user.

### DIFF
--- a/lampod-cli/src/main.rs
+++ b/lampod-cli/src/main.rs
@@ -168,9 +168,20 @@ fn run(args: LampoCliArgs) -> error::Result<()> {
             wallet
         }
     } else {
-        // If there is a file, we tell the user to use --restore-wallet
+        // If there is a file, we load the wallet with a warning
         if Path::new(&format!("{}/wallet.dat", words_path)).exists() {
-            error::bail!("Wallet already exists. Please use --restore-wallet")
+            // Load the mnemonic from the file
+            log::warn!("Loading from existing wallet");
+            let mnemonic = load_words_from_file(format!("{}/wallet.dat", words_path))?;
+            let wallet = match client.kind() {
+                lampo_common::backend::BackendKind::Core => {
+                    CoreWalletManager::restore(Arc::new(lampo_conf.clone()), &mnemonic)?
+                }
+                lampo_common::backend::BackendKind::Nakamoto => {
+                    error::bail!("wallet is not implemented for nakamoto")
+                }
+            };
+            wallet
         } else {
             let (wallet, mnemonic) = match client.kind() {
                 lampo_common::backend::BackendKind::Core => {
--- a/lampod-cli/src/main.rs
+++ b/lampod-cli/src/main.rs
@@ -8,6 +8,10 @@ use std::str::FromStr;
 use std::sync::Arc;
 use std::thread::JoinHandle;
 
+use std::fs::{File, OpenOptions};
+use std::io::{Read, Write};
+use std::path::Path;
+
 use radicle_term as term;
 
 use lampo_bitcoind::BitcoinCore;
@@ -44,21 +48,39 @@ fn main() -> error::Result<()> {
     Ok(())
 }
 
+fn write_words_to_file<P: AsRef<Path>>(path: P, words: String) -> error::Result<()> {
+    let mut file = OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .open(path)?;
+
+    // FIXME: we should give the possibility to encrypt this file.
+    file.write_all(words.as_bytes())?;
+    Ok(())
+}
+
+fn load_words_from_file<P: AsRef<Path>>(path: P) -> error::Result<String> {
+    let mut file = File::open(path.as_ref())?;
+    let mut content = String::new();
+
+    file.read_to_string(&mut content)?;
+
+    if content.is_empty() {
+        let path = path.as_ref().to_string_lossy().to_string();
+        error::bail!("The content of the wallet located at `{path}`. You lost the secret? Please report a bug this should never happens")
+    } else {
+        Ok(content)
+    }
+}
+
 /// Return the root directory.
 fn run(args: LampoCliArgs) -> error::Result<()> {
-    let mnemonic = if args.restore_wallet {
-        let inputs: String = term::input(
-            "BIP 39 Mnemonic",
-            None,
-            Some("To restore the wallet, lampo needs a BIP39 mnemonic with words separated by spaces."),
-        )?;
-        Some(inputs)
-    } else {
-        None
-    };
+    let restore_wallet = args.restore_wallet;
 
     // After this point the configuration is ready!
     let mut lampo_conf: LampoConf = args.try_into()?;
+
     log::debug!(target: "lampod-cli", "init wallet ..");
     // init the logger here
     logger::init(
@@ -97,36 +119,73 @@ fn run(args: LampoCliArgs) -> error::Result<()> {
         _ => error::bail!("client {:?} not supported", client),
     };
 
-    let wallet = if let Some(ref _private_key) = lampo_conf.private_key {
-        unimplemented!()
-    } else if mnemonic.is_none() {
-        let (wallet, mnemonic) = match client.kind() {
-            lampo_common::backend::BackendKind::Core => {
-                CoreWalletManager::new(Arc::new(lampo_conf.clone()))?
-            }
-            lampo_common::backend::BackendKind::Nakamoto => {
-                error::bail!("wallet is not implemented for nakamoto")
-            }
-        };
+    if let Some(ref _private_key) = lampo_conf.private_key {
+        error::bail!("Option to force a private key not available at the moment")
+    }
 
-        radicle_term::success!("Wallet Generated, please store these words in a safe way");
-        radicle_term::println(
-            radicle_term::format::badge_primary("wallet-keys"),
-            format!("{}", radicle_term::format::highlight(mnemonic)),
-        );
-        wallet
+    let words_path = format!("{}/", lampo_conf.path());
+    // There are several case in this if-else, that are:
+    // 1. lampo is running on a fresh os:
+    //   1.1: the user has a wallet, so need to specify `--restore-wallet`
+    //   1.2: the user does not have a wallet so lampo should generate a new seed for it and store it in a file.
+    // 2. lampo is running on os where there is a wallet, lampo will load the seeds from wallet:
+    //   2.1: The user keep specify --restore-wallet, so lampo should return an error an tell the user that there is already a wallet
+    //   2.2: The user do not specify the --restore-wallet, so lampo load from disk the file, if there is no file it return an error
+    // FIXME: there is a problem of code duplication here, we should move this code in utils functions.
+    let wallet = if restore_wallet {
+        if Path::new(&format!("{}/wallet.dat", words_path)).exists() {
+            // Load the mnemonic from the file
+            let mnemonic = load_words_from_file(format!("{}/wallet.dat", words_path))?;
+            let wallet = match client.kind() {
+                lampo_common::backend::BackendKind::Core => {
+                    CoreWalletManager::restore(Arc::new(lampo_conf.clone()), &mnemonic)?
+                }
+                lampo_common::backend::BackendKind::Nakamoto => {
+                    error::bail!("wallet is not implemented for nakamoto")
+                }
+            };
+            wallet
+        } else {
+            // If file doesn't exist, ask for user input
+            let mnemonic: String = term::input(
+                "BIP 39 Mnemonic",
+                None,
+                Some("To restore the wallet, lampo needs the BIP39 mnemonic with words separated by spaces."),
+            )?;
+            // FIXME: make some sanity check about the mnemonic string
+            let wallet = match client.kind() {
+                lampo_common::backend::BackendKind::Core => {
+                    // SAFETY: It is safe to unwrap the mnemonic because we check it
+                    // before.
+                    CoreWalletManager::restore(Arc::new(lampo_conf.clone()), &mnemonic)?
+                }
+                lampo_common::backend::BackendKind::Nakamoto => {
+                    error::bail!("wallet is not implemented for nakamoto")
+                }
+            };
+
+            write_words_to_file(format!("{}/wallet.dat", words_path), mnemonic)?;
+            wallet
+        }
     } else {
-        match client.kind() {
-            lampo_common::backend::BackendKind::Core => {
-                // SAFETY: It is safe to unwrap the mnemonic because we check it
-                // before.
-                CoreWalletManager::restore(Arc::new(lampo_conf.clone()), &mnemonic.unwrap())?
-            }
-            lampo_common::backend::BackendKind::Nakamoto => {
-                error::bail!("wallet is not implemented for nakamoto")
-            }
+        // If there is a file, we tell the user to use --restore-wallet
+        if Path::new(&format!("{}/wallet.dat", words_path)).exists() {
+            error::bail!("Wallet already exists. Please use --restore-wallet")
+        } else {
+            let (wallet, mnemonic) = match client.kind() {
+                lampo_common::backend::BackendKind::Core => {
+                    CoreWalletManager::new(Arc::new(lampo_conf.clone()))?
+                }
+                lampo_common::backend::BackendKind::Nakamoto => {
+                    error::bail!("wallet is not implemented for nakamoto")
+                }
+            };
+
+            write_words_to_file(format!("{}/wallet.dat", words_path), mnemonic)?;
+            wallet
         }
     };
+
     log::debug!(target: "lampod-cli", "wallet created with success");
     let mut lampod = LampoDaemon::new(lampo_conf.clone(), Arc::new(wallet));
 


lampod-cli/src/main.rs
```diff
@@ -44,15 +48,57 @@ fn main() -> error::Result<()> {
     Ok(())
 }
 
+fn write_words_to_file<P: AsRef<Path>>(path: P, words: &Option<String>) -> io::Result<()> {
```
vincenzopalazzo, line 51:
```suggestion
fn write_words_to_file<P: AsRef<Path>>(path: P, words: Option<String>) -> io::Result<()> {
```

Probably better?

lampod-cli/src/main.rs
```diff
@@ -44,15 +48,57 @@ fn main() -> error::Result<()> {
...
+        .truncate(true)
+        .open(path)?;
+
+    match words {
+        Some(ref value) => {
+            file.write_all(value.as_bytes())?;
+        }
+        None => {
+            file.write_all(b"")?;
+        }
```
vincenzopalazzo, line 64: I am not able to understand this, why ?

lampod-cli/src/main.rs
```diff
@@ -44,15 +48,57 @@ fn main() -> error::Result<()> {
...
+    if content.is_empty() {
+        Ok(None)
+    } else {
+        Ok(Some(content))
+    }
+}
+
 /// Return the root directory.
 fn run(args: LampoCliArgs) -> error::Result<()> {
+    let path = std::env::home_dir().expect("Impossible to get the home directory");
```
vincenzopalazzo, line 85:
```suggestion
    let path = std::env::home_dir().ok_or(error::anyhow!("Impossible to get the home directory"))?;
```

In lampo we never panic

lampod-cli/src/main.rs
```diff
@@ -44,15 +48,57 @@ fn main() -> error::Result<()> {
...
+    let path = path.to_str().unwrap();
+    let words_path = format!("{}/.lampo/{}", path, args.network.as_ref().unwrap());
```
vincenzopalazzo, line 87: we may have the lampo home that include the network somewhere else? probably the LampoConf? or similar?

lampod-cli/src/main.rs
```diff
@@ -44,15 +48,57 @@ fn main() -> error::Result<()> {
...
+        if Path::new(&format!("{}/wallet.dat", words_path)).exists() {
+            // Load the mnemonic from the file
+            load_words_from_file(format!("{}/wallet.dat", words_path))?
+        } else {
+            // If file doesn't exist, ask for user input
+            let inputs: String = term::input(
+                "BIP 39 Mnemonic",
+                None,
+                Some("To restore the wallet, lampo needs the BIP39 mnemonic with words separated by spaces."),
+            )?;
```
vincenzopalazzo, line 99: <3 
```suggestion
            let inputs: String = term::input(
                "BIP 39 Mnemonic",
                None,
                Some("lampo needs the BIP39 mnemonic with words separated by spaces in order to create restore the wallet."),
            )?;
```

lampod-cli/src/main.rs
```diff
@@ -112,15 +158,19 @@ fn run(args: LampoCliArgs) -> error::Result<()> {
         radicle_term::success!("Wallet Generated, please store these words in a safe way");
         radicle_term::println(
             radicle_term::format::badge_primary("wallet-keys"),
-            format!("{}", radicle_term::format::highlight(mnemonic)),
+            format!("{}", radicle_term::format::highlight(mnemonic.clone())),
```
vincenzopalazzo, line 161:
```suggestion
            format!("{}", radicle_term::format::highlight(&mnemonic)),
```

What about this? we can avoid to clone it?

lampod-cli/src/main.rs
```diff
@@ -112,15 +158,19 @@ fn run(args: LampoCliArgs) -> error::Result<()> {
...
         );
+        write_words_to_file(format!("{}/wallet.dat", words_path), &Some(mnemonic))?;
```
vincenzopalazzo, line 163:
```suggestion
        write_words_to_file(format!("{}/wallet.dat", words_path), Some(mnemonic))?;
```

lampod-cli/src/main.rs
```diff
@@ -112,15 +158,19 @@ fn run(args: LampoCliArgs) -> error::Result<()> {
...
         wallet
     } else {
         match client.kind() {
             lampo_common::backend::BackendKind::Core => {
                 // SAFETY: It is safe to unwrap the mnemonic because we check it
                 // before.
-                CoreWalletManager::restore(Arc::new(lampo_conf.clone()), &mnemonic.unwrap())?
+                CoreWalletManager::restore(
+                    Arc::new(lampo_conf.clone()),
+                    &mnemonic.clone().unwrap(),
```
vincenzopalazzo, line 172:
```suggestion
                    &mnemonic.clone().unwrap(),
```

Why we are calling `unwrap()`?

lampod-cli/src/main.rs
Psycho-Pirate: There is a directory in lampo conf but we might have to clone the `args` as `args` are getting borrowed while creating lampo conf. If we generate lampo conf before we get error in this line.
`let mnemonic = if args.restore_wallet {`

lampod-cli/src/main.rs
Psycho-Pirate: because mnemonic is of type `Option<string>` here

lampod-cli/src/main.rs
vincenzopalazzo: OK let think more about this!

What is the case where the mnemonic is None?

lampod-cli/src/main.rs
vincenzopalazzo: You can take the reference of the path before the `let-if`?  If you provide an example I could help you to fix this problem

lampod-cli/src/main.rs
vincenzopalazzo: Mh, but the mnemonic should be always specified at this point right? otherwise lampo crash.

So there is a way to set the mnemonic to `String` and not `Option`?

lampod-cli/src/main.rs
Psycho-Pirate: Yes, ususally this case wouldnt happen. This will only happen if the mnemonic is none, maybe the user doesnt enter anything when the mnemonic is asked.

lampod-cli/src/main.rs
Psycho-Pirate: This can be implement. But we do have a comment there already

```
lampo_common::backend::BackendKind::Core => {
                // SAFETY: It is safe to unwrap the mnemonic because we check it
                // before.
                CoreWalletManager::restore(
                    Arc::new(lampo_conf.clone()),
                    &mnemonic.clone().unwrap(),
                )?
            }
```

lampod-cli/src/main.rs
Psycho-Pirate: I have made a fix. It should be fine now.

lampod-cli/src/main.rs
vincenzopalazzo: We should return an error if the user does not pass the correct mnemonic!

Your code is writing an invalid seed, which in some cases can lead to lost funds. This task that you are doing may looks trivial but it is crucial for user funds, so please pay attention on the validation

lampod-cli/src/main.rs
```diff
@@ -44,21 +48,61 @@ fn main() -> error::Result<()> {
...
+            load_words_from_file(format!("{}/wallet.dat", words_path))?
+        } else {
+            // If file doesn't exist, ask for user input
+            let inputs: String = term::input(
+                "BIP 39 Mnemonic",
+                None,
+                Some("To restore the wallet, lampo needs the BIP39 mnemonic with words separated by spaces."),
+            )?;
+            Some(inputs)
+        }
```
vincenzopalazzo, line 101:
```suggestion
        if Path::new(&format!("{}/wallet.dat", words_path)).exists() {
            // Load the mnemonic from the file
            load_words_from_file(format!("{}/wallet.dat", words_path))?
        } else {
            // If file doesn't exist, ask for user input
            let inputs: String = term::input(
                "BIP 39 Mnemonic",
                None,
                Some("To restore the wallet, lampo needs the BIP39 mnemonic with words separated by spaces."),
            )?;
            inputs
        }
```

What is the reason to exit from this if with a Some? you know that will be always some. So at this point the mnemonic should be known.

Then in case we are not restoring the wallet, we need to do two things:

1. The file exist? if yes return an error to the user, otherwise ldk will messup with the node secrets and you are losing funds
2. if the wallet file do not exist, generate the mnemonic. See a couple of line down `CoreWalletManager::new(Arc::new(lampo_conf.clone()))?`  probably it is time to refactor this code

lampod-cli/src/main.rs
vincenzopalazzo: Yeah this could works but the code will be not sound, there is no reason to arrived here with the mnemonic still of type `Option`. See https://github.com/vincenzopalazzo/lampo.rs/pull/295#discussion_r1823144469

lampod-cli/src/main.rs
```diff
@@ -44,21 +48,61 @@ fn main() -> error::Result<()> {
...
-            None,
-            Some("To restore the wallet, lampo needs a BIP39 mnemonic with words separated by spaces."),
-        )?;
-        Some(inputs)
+    let restore_wallet = args.restore_wallet;
+
+    // After this point the configuration is ready!
+    let mut lampo_conf: LampoConf = args.try_into()?;
+
+    let words_path = format!("{}/", lampo_conf.path());
```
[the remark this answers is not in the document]
Psycho-Pirate: Makes sense! I will create a follow up PR for this

lampod-cli/src/main.rs
```diff
@@ -44,21 +48,61 @@ fn main() -> error::Result<()> {
...
+            let inputs: String = term::input(
+                "BIP 39 Mnemonic",
+                None,
+                Some("To restore the wallet, lampo needs the BIP39 mnemonic with words separated by spaces."),
+            )?;
+            Some(inputs)
+        }
     } else {
         None
     };
```
Psycho-Pirate, line 104: How is this looking?

```suggestion
    let mnemonic = if restore_wallet {
        if Path::new(&format!("{}/wallet.dat", words_path)).exists() {
            // Load the mnemonic from the file
            load_words_from_file(format!("{}/wallet.dat", words_path))?
        } else {
            // If file doesn't exist, ask for user input
            let inputs: String = term::input(
                "BIP 39 Mnemonic",
                None,
                Some("To restore the wallet, lampo needs the BIP39 mnemonic with words separated by spaces."),
            )?;
            Some(inputs)
        }
    } else {
        if Path::new(&format!("{}/wallet.dat", words_path)).exists() {
            // if file exist and wallet is not restored, return error
            error::bail!("Wallet already exists. Please use --restore-wallet")
        } else {
            None
        }
    };
```

lampod-cli/src/main.rs
Psycho-Pirate: Okay

lampod-cli/src/main.rs
vincenzopalazzo: >The advantage of using some here is that it makes mnemonic an Option type. We need it to be option because we also have a none case, the case in which restore wallet is not called.

The mnemonic to generate the wallet must be always define, I do not understand where the None case is helping you?

lampod-cli/src/main.rs
vincenzopalazzo: Mh not sure can you explaing what you want to do with this if else code?

Psycho-Pirate: The else case is used when lampod command is called without `--restore-wallet`. Here it checks if the wallet file exists or not. If it does it tells the user to use restore wallet. If it doesnt it will go ahead and mark the mnemonic as None.

Later this is checked and if mnemonic is none, new words are generated.

lampod-cli/src/main.rs
Psycho-Pirate: The none case is later checked to generate new words here. https://github.com/Psycho-Pirate/lampo.rs/blob/fbdf05798fad7ab7a264cb33c04398a847bca55d/lampod-cli/src/main.rs#L146

If mnemonic is set to none, it means that we are having a new lampo session. Hence new words are generated.

Harshit933:
There is some problem with the wallet design as of now. If there is a wallet that already exists in the `.lampo/regtest/wallet.dat` and if we accidently run `lampod-cli --network <chain>` the phrase in the `wallet.dat` file is replaced with a freshly generated one [This is a major issue], maybe we can check if the `wallet.dat` file already exists then don't proceed with generating new phrases, could be related to https://github.com/vincenzopalazzo/lampo.rs/issues/296. Also, I think we should password protect the file and write only the encrypted seed phrase in the `wallet.dat` file but it could be a work for another PR.

lampod-cli/src/main.rs
Harshit933: This is better.

vincenzopalazzo:
I played with the code, what is wrong with my solution? @Psycho-Pirate 

It solved some problems and removed the mnemonic as a `Option` type because there was some problem with the write and read the file.

However, I am open to discussing if there is any case that I am not keeping into count?

Probably this is also the problem that @Harshit933 point out inside the review 

P.S: At this stage of lampo we should refactor the code and make it better, and not try to work with the current codebase, because most of the time there is no optimal code.

----

The best option to see the diff in action is to use `git apply <file>`



```diff
diff --git a/lampod-cli/src/main.rs b/lampod-cli/src/main.rs
index 5864816..c9a2d8f 100644
--- a/lampod-cli/src/main.rs
+++ b/lampod-cli/src/main.rs
@@ -48,33 +48,29 @@ fn main() -> error::Result<()> {
     Ok(())
 }

-fn write_words_to_file<P: AsRef<Path>>(path: P, words: Option<String>) -> io::Result<()> {
+fn write_words_to_file<P: AsRef<Path>>(path: P, words: String) -> error::Result<()> {
     let mut file = OpenOptions::new()
         .write(true)
         .create(true)
         .truncate(true)
         .open(path)?;

-    match words {
-        Some(ref value) => {
-            file.write_all(value.as_bytes())?;
-        }
-        None => {}
-    }
-
+    // FIXME: we should give the possibility to encrypt this file.
+    file.write_all(words.as_bytes())?;
     Ok(())
 }

-fn load_words_from_file<P: AsRef<Path>>(path: P) -> io::Result<Option<String>> {
-    let mut file = File::open(path)?;
+fn load_words_from_file<P: AsRef<Path>>(path: P) -> error::Result<String> {
+    let mut file = File::open(path.as_ref())?;
     let mut content = String::new();

     file.read_to_string(&mut content)?;

     if content.is_empty() {
-        Ok(None)
+        let path = path.as_ref().to_string_lossy().to_string();
+        error::bail!("The content of the wallet located at `{path}`. You lost the secret? Please report a bug this should never happens")
     } else {
-        Ok(Some(content))
+        Ok(content)
     }
 }

@@ -85,24 +81,6 @@ fn run(args: LampoCliArgs) -> error::Result<()> {
     // After this point the configuration is ready!
     let mut lampo_conf: LampoConf = args.try_into()?;

-    let words_path = format!("{}/", lampo_conf.path());
-    let mnemonic = if restore_wallet {
-        if Path::new(&format!("{}/wallet.dat", words_path)).exists() {
-            // Load the mnemonic from the file
-            load_words_from_file(format!("{}/wallet.dat", words_path))?
-        } else {
-            // If file doesn't exist, ask for user input
-            let inputs: String = term::input(
-                "BIP 39 Mnemonic",
-                None,
-                Some("To restore the wallet, lampo needs the BIP39 mnemonic with words separated by spaces."),
-            )?;
-            Some(inputs)
-        }
-    } else {
-        None
-    };
-
     log::debug!(target: "lampod-cli", "init wallet ..");
     // init the logger here
     logger::init(
@@ -141,47 +119,90 @@ fn run(args: LampoCliArgs) -> error::Result<()> {
         _ => error::bail!("client {:?} not supported", client),
     };

-    let wallet = if let Some(ref _private_key) = lampo_conf.private_key {
-        unimplemented!()
-    } else if mnemonic.is_none() {
-        let (wallet, mnemonic) = match client.kind() {
-            lampo_common::backend::BackendKind::Core => {
-                CoreWalletManager::new(Arc::new(lampo_conf.clone()))?
-            }
-            lampo_common::backend::BackendKind::Nakamoto => {
-                error::bail!("wallet is not implemented for nakamoto")
-            }
-        };
-
-        radicle_term::success!("Wallet Generated, please store these words in a safe way");
-        radicle_term::println(
-            radicle_term::format::badge_primary("wallet-keys"),
-            format!("{}", radicle_term::format::highlight(&mnemonic)),
-        );
-        write_words_to_file(format!("{}/wallet.dat", words_path), Some(mnemonic))?;
-        wallet
+    if let Some(ref _private_key) = lampo_conf.private_key {
+        error::bail!("Option to force a private key not available at the moment")
+    }
+
+    let words_path = format!("{}/", lampo_conf.path());
+    // There are several case in this if-else, that are:
+    // 1. lampo is running on a fresh os:
+    //   1.1: the user has a wallet, so need to specify `--restore-wallet`
+    //   1.2: the user does not have a wallet so lampo should generate a new seed for it and store it in a file.
+    // 2. lampo is running on os where there is a wallet, lampo will load the seeds from wallet:
+    //   2.1: The user keep specify --restore-wallet, so lampo should return an error an tell the user that there is already a wallet
+    //   2.2: The user do not specify the --restore-wallet, so lampo load from disk the file, if there is no file it return an error
+    // FIXME: there is a problem of code duplication here, we should move this code in utils functions.
+    let wallet = if restore_wallet {
+        if Path::new(&format!("{}/wallet.dat", words_path)).exists() {
+            // Load the mnemonic from the file
+            let mnemonic = load_words_from_file(format!("{}/wallet.dat", words_path))?;
+            let wallet = match client.kind() {
+                lampo_common::backend::BackendKind::Core => {
+                    // SAFETY: It is safe to unwrap the mnemonic because we check it
+                    // before.
+                    CoreWalletManager::restore(Arc::new(lampo_conf.clone()), &mnemonic)?
+                }
+                lampo_common::backend::BackendKind::Nakamoto => {
+                    error::bail!("wallet is not implemented for nakamoto")
+                }
+            };
+            wallet
+        } else {
+            // If file doesn't exist, ask for user input
+            let mnemonic: String = term::input(
+                "BIP 39 Mnemonic",
+                None,
+                Some("To restore the wallet, lampo needs the BIP39 mnemonic with words separated by spaces."),
+            )?;
+            // FIXME: make some sanity check about the mnemonic string
+            let wallet = match client.kind() {
+                lampo_common::backend::BackendKind::Core => {
+                    // SAFETY: It is safe to unwrap the mnemonic because we check it
+                    // before.
+                    CoreWalletManager::restore(Arc::new(lampo_conf.clone()), &mnemonic)?
+                }
+                lampo_common::backend::BackendKind::Nakamoto => {
+                    error::bail!("wallet is not implemented for nakamoto")
+                }
+            };
+
+            write_words_to_file(format!("{}/wallet.dat", words_path), mnemonic)?;
+            wallet
+        }
     } else {
-        match client.kind() {
-            lampo_common::backend::BackendKind::Core => {
-                // SAFETY: It is safe to unwrap the mnemonic because we check it
-                // before.
-                CoreWalletManager::restore(
-                    Arc::new(lampo_conf.clone()),
-                    &mnemonic.clone().unwrap(),
-                )?
-            }
-            lampo_common::backend::BackendKind::Nakamoto => {
-                error::bail!("wallet is not implemented for nakamoto")
-            }
+        // There is a file, so we load it, and we tell the user through logs that we found it
+        if Path::new(&format!("{}/wallet.dat", words_path)).exists() {
+            // Load the mnemonic from the file
+            let mnemonic = load_words_from_file(format!("{}/wallet.dat", words_path))?;
+            let wallet = match client.kind() {
+                lampo_common::backend::BackendKind::Core => {
+                    // SAFETY: It is safe to unwrap the mnemonic because we check it
+                    // before.
+                    CoreWalletManager::restore(Arc::new(lampo_conf.clone()), &mnemonic)?
+                }
+                lampo_common::backend::BackendKind::Nakamoto => {
+                    error::bail!("wallet is not implemented for nakamoto")
+                }
+            };
+            wallet
+        } else {
+            let (wallet, mnemonic) = match client.kind() {
+                lampo_common::backend::BackendKind::Core => {
+                    CoreWalletManager::new(Arc::new(lampo_conf.clone()))?
+                }
+                lampo_common::backend::BackendKind::Nakamoto => {
+                    error::bail!("wallet is not implemented for nakamoto")
+                }
+            };
+
+            write_words_to_file(format!("{}/wallet.dat", words_path), mnemonic)?;
+            wallet
         }
     };
+
     log::debug!(target: "lampod-cli", "wallet created with success");
     let mut lampod = LampoDaemon::new(lampo_conf.clone(), Arc::new(wallet));

-    if !Path::new(&format!("{}/wallet.dat", words_path)).exists() {
-        write_words_to_file(format!("{}/wallet.dat", words_path), mnemonic)?;
-    }
-
     // Init the lampod
     lampod.init(client)?;
```

lampod-cli/src/main.rs
vincenzopalazzo: What about https://github.com/vincenzopalazzo/lampo.rs/pull/295#issuecomment-2453061580 ?

lampod-cli/src/main.rs
vincenzopalazzo: Naa you are overcomplicating it or trusting too much the lampo shitty code

What do you think about https://github.com/vincenzopalazzo/lampo.rs/pull/295#issuecomment-2453061580 ?

Harshit933:
@vincenzopalazzo Looks good to me.

Psycho-Pirate:
> I played with the code, what is wrong with my solution? @Psycho-Pirate
> 
> It solved some problems and removed the mnemonic as a `Option` type because there was some problem with the write and read the file.
> 
> However, I am open to discussing if there is any case that I am not keeping into count?
> 
> Probably this is also the problem that @Harshit933 point out inside the review
> 
> P.S: At this stage of lampo we should refactor the code and make it better, and not try to work with the current codebase, because most of the time there is no optimal code.
> 

This looks good and covers everything. We can go forward with this approach. But I have some suggestions:
1. When `restore_wallet` is false, we are checking for the file and loading the wallet if file exists. I think here we should exit with error saying something like "Wallet already exists. Please use --restore-wallet".

2. In case we are generating a new wallet we should also display the words generated as we were doing in the existing code.

vincenzopalazzo:
>When restore_wallet is false, we are checking for the file and loading the wallet if file exists. I think here we should exit with error saying something like "Wallet already exists. Please use --restore-wallet".

This sounds good to me, thanks for the suggestion.

>In case we are generating a new wallet we should also display the words generated as we were doing in the existing code.

This is a privacy leak and we should avoid it. Then we should add a command in `lampod-cli show-bip39` that return the words back, what do you think about this?

Psycho-Pirate:
> 
> This is a privacy leak and we should avoid it. Then we should add a command in `lampod-cli show-bip39` that return the words back, what do you think about this?

I agree! So I will go ahead with this approach and I will add the command `lampod-cli show-bip39` in a followup PR

lampod-cli/src/main.rs
```diff
@@ -141,47 +119,78 @@ fn run(args: LampoCliArgs) -> error::Result<()> {
...
+    //   2.2: The user do not specify the --restore-wallet, so lampo load from disk the file, if there is no file it return an error
+    // FIXME: there is a problem of code duplication here, we should move this code in utils functions.
+    let wallet = if restore_wallet {
+        if Path::new(&format!("{}/wallet.dat", words_path)).exists() {
+            // Load the mnemonic from the file
+            let mnemonic = load_words_from_file(format!("{}/wallet.dat", words_path))?;
+            let wallet = match client.kind() {
+                lampo_common::backend::BackendKind::Core => {
+                    // SAFETY: It is safe to unwrap the mnemonic because we check it
+                    // before.
```
vincenzopalazzo, line 142:
```suggestion
```

there is no unwrap anymore

lampod-cli/src/main.rs
```diff
@@ -141,47 +119,78 @@ fn run(args: LampoCliArgs) -> error::Result<()> {
...
+                    CoreWalletManager::restore(Arc::new(lampo_conf.clone()), &mnemonic)?
+                }
+                lampo_common::backend::BackendKind::Nakamoto => {
```
vincenzopalazzo, line 145:
```suggestion
                lampo_common::backend::BackendKind::Nakamoto => {
```

Idea for a follow-up PR, we should remove nakamoto from here and from all the project

lampod-cli/src/main.rs
```diff
@@ -141,47 +119,78 @@ fn run(args: LampoCliArgs) -> error::Result<()> {
...
+                Some("To restore the wallet, lampo needs the BIP39 mnemonic with words separated by spaces."),
+            )?;
+            // FIXME: make some sanity check about the mnemonic string
+            let wallet = match client.kind() {
+                lampo_common::backend::BackendKind::Core => {
+                    // SAFETY: It is safe to unwrap the mnemonic because we check it
+                    // before.
+                    CoreWalletManager::restore(Arc::new(lampo_conf.clone()), &mnemonic)?
+                }
+                lampo_common::backend::BackendKind::Nakamoto => {
```
vincenzopalazzo, line 164: same consideration here for a folloup PR

lampod-cli/src/main.rs
```diff
@@ -141,47 +119,78 @@ fn run(args: LampoCliArgs) -> error::Result<()> {
...
-                    Arc::new(lampo_conf.clone()),
-                    &mnemonic.clone().unwrap(),
-                )?
-            }
-            lampo_common::backend::BackendKind::Nakamoto => {
-                error::bail!("wallet is not implemented for nakamoto")
-            }
+        // If there is a file, we tell the user to use --restore-wallet
+        if Path::new(&format!("{}/wallet.dat", words_path)).exists() {
+            error::bail!("Wallet already exists. Please use --restore-wallet")
```
vincenzopalazzo, line 175: nit: I reconsider my self here. This make scripting or using `systemd` in linux much harder because it is not that trivial run two flavor of command.

Probably you can print a warning message here? and say that lampo is restarting an existing wallet?

vincenzopalazzo (changes requested):
Ok two comment and we are ready to go, sorry about the confusion here anyway!

lampod-cli/src/main.rs
Psycho-Pirate: Maybe we should! Do other implementations support Nakamoto? If they do then we should keep it to make Lampo more versatile.

lampod-cli/src/main.rs
Psycho-Pirate: Okay

lampod-cli/src/main.rs
vincenzopalazzo: I am the current maintainer of nakamo. Core lightning supports it but we should drop it in a followup PR (IIRC I already drop `lampo-nakamoto`)

vincenzopalazzo: I am the current maintainer of nakamo. Core lightning supports it but we should drop it in a followup PR (IIRC I already drop lampo-nakamoto)

lampod-cli/src/main.rs
```diff
@@ -168,9 +168,20 @@ fn run(args: LampoCliArgs) -> error::Result<()> {
             wallet
         }
     } else {
-        // If there is a file, we tell the user to use --restore-wallet
+        // If there is a file, we load the wallet with a warning
         if Path::new(&format!("{}/wallet.dat", words_path)).exists() {
-            error::bail!("Wallet already exists. Please use --restore-wallet")
+            // Load the mnemonic from the file
+            radicle_term::warning("Loading from existing wallet");
```
vincenzopalazzo, line 174: No,  use `log::warn("Loading from existing wallet");`

vincenzopalazzo (changes requested):
Another iteration and probably we are good

vincenzopalazzo:
@Psycho-Pirate when you are ready please reassign my review but I think I self review that you do yourself on the code would be good too.

In addition, we need to work on how you present the git history (6 comment for a `git apply <patch>` that I give are too many https://github.com/vincenzopalazzo/lampo.rs/pull/295/commits)

I think this is a good reading https://cbea.ms/git-commit/ and this is a bad example of commit history that cause a close of the PR https://github.com/vincenzopalazzo/lampo.rs/pull/284#issuecomment-2380571525

vincenzopalazzo:
>Should I also squash the commits into one?

Up to you, the history should make sense and be easy to cherry-pick, probably this is a good reading https://www.kernel.org/doc/html/v4.10/process/submitting-patches.html#separate-your-changes

Psycho-Pirate:
Squashed it to two commits for cleaner commit history

vincenzopalazzo (approved):
ACK aca1633cac69fd7758cc98011f153db472b2a118

Assuming that you tested out 😸

PR #361, opened 2025-02-18
feat: added clap for cli

### LINKED ISSUE #357: using clap for command line parsing inside the lampo-cli
We discussed a while back with @chrisguida where I did not want to add the clap crate because was too big and I could rewrite it in a less dependencies.

However, I did not consider that time was a problem and I could not rewrite everything, so this issue is suggesting to support clap inside the `lampo-cli`

An example of how to use clap is https://github.com/coffee-tools/coffee/blob/master/coffee_cmd/src/cmd.rs or/and https://github.com/lightningdevkit/ldk-server/blob/main/ldk-server-cli/src/main.rs

I suggest building on the next branch https://github.com/vincenzopalazzo/lampo.rs/tree/next which contains the more recent version of the code base.

Gmin2:
image

Fixes #357

### DIFF
--- a/lampo-cli/Cargo.toml
+++ b/lampo-cli/Cargo.toml
@@ -9,3 +9,4 @@ lampo-client = { path = "../lampo-client" }
 log = { version = "0.4", features = ["std"] }
 lexopt = { version = "0.3" }
 radicle-term = { git = "https://github.com/radicle-dev/heartwood.git" }
+clap = { version = "4.5", features = ["derive"] }
--- a/lampo-cli/src/args.rs
+++ b/lampo-cli/src/args.rs
@@ -1,163 +1,182 @@
-use std::collections::HashMap;
-
-use radicle_term as term;
-
+use clap::{ArgAction, Parser, Subcommand};
 use lampo_common::error;
 use lampo_common::json;
+use std::collections::HashMap;
 
-#[derive(Debug)]
+#[derive(Parser, Debug)]
+#[command(
+    name = "lampod-cli",
+    about = "Lampo Daemon command line",
+    version,
+    long_about = None
+)]
+pub struct Cli {
+    #[arg(short = 'd', long = "data-dir")]
+    pub data_dir: Option<String>,
+
+    #[arg(short = 'n', long, default_value = "testnet")]
+    pub network: String,
+
+    #[arg(short, long)]
+    pub socket: Option<String>,
+
+    #[command(subcommand)]
+    pub command: Commands,
+}
+
+#[derive(Subcommand, Debug)]
+pub enum Commands {
+    GetInfo,
+    ListChannels,
+    OpenChannel {
+        #[arg(short, long)]
+        node_id: String,
+
+        #[arg(long)]
+        capacity: u64,
+
+        #[arg(long)]
+        push_msat: Option<u64>,
+
+        #[arg(long, action = ArgAction::SetTrue)]
+        announce: Option<bool>,
+    },
+
+    Connect {
+        #[arg(short, long)]
+        node_id: String,
+        #[arg(short, long)]
+        addr: String,
+    },
+
+    CreateInvoice {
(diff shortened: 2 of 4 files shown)


lampo-cli/src/args.rs
```diff
@@ -1,163 +1,182 @@
...
+        Commands::GetInfo => ("getinfo".to_string(), HashMap::new()),
+
+        Commands::ListChannels => ("listchannels".to_string(), HashMap::new()),
+
+        Commands::OpenChannel {
+            node_id,
+            capacity,
+            push_msat,
+            announce,
+        } => {
```
vincenzopalazzo, line 114: we can remove the spaces?
```suggestion
    let (method, args) = match cli.command {
        Commands::GetInfo => ("getinfo".to_string(), HashMap::new()),
        Commands::ListChannels => ("listchannels".to_string(), HashMap::new()),
        Commands::OpenChannel {
            node_id,
            capacity,
            push_msat,
            announce,
        } => {
```

lampo-cli/src/args.rs
```diff
@@ -1,163 +1,182 @@
...
-                            let val = json::json!(val);
-                            args.insert(key.clone(), val);
-                        } else {
-                            let val = json::json!(val);
-                            args.insert(key, val);
-                        }
+            ("payinvoice".to_string(), args)
+        }
+
+        Commands::Call { method, params } => {
```
vincenzopalazzo, line 160: probably we can remove this `Commands::Call` for now I feel that this is overcomplicating

lampo-cli/src/main.rs
```diff
@@ -1,24 +1,23 @@
...
-use std::process::exit;
-
-use radicle_term as term;
-
 use lampo_client::errors::Error;
 use lampo_client::UnixClient;
 use lampo_common::error;
 use lampo_common::json;
+use radicle_term as term;
+use std::process::exit;
```
vincenzopalazzo, line 8: We have an order of the imports if you will pay attention to is you will not the following orders

```
<std imports>

<external dependencies>

<lampo imports>
```

vincenzopalazzo (changes requested):
Thanks for working on this, I feel that this will take some iteration to find the right solution of us but your code looks good to me for now. Thanks for working on it :)

lampo-cli/Cargo.toml
```diff
@@ -9,3 +9,4 @@ lampo-client = { path = "../lampo-client" }
 log = { version = "0.4", features = ["std"] }
 lexopt = { version = "0.3" }
```
vincenzopalazzo, line 10:
```suggestion
```

We can remove this now that we are using clap

Harshit933 (changes requested):
Haven't tested the code for now, but is in a right direction

lampo-cli/src/args.rs
```diff
@@ -1,163 +1,182 @@
...
+    version,
+    long_about = None
+)]
+pub struct Cli {
+    #[arg(short = 'd', long = "data-dir")]
+    pub data_dir: Option<String>,
 
-use lampo_common::error;
-use lampo_common::json;
+    #[arg(short = 'n', long, default_value = "testnet")]
```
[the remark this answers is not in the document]
vincenzopalazzo: >Why is the default value testnet here?

probably because it was in this way before! thinking if we should default to signet?

In any way we should not default this to `bitcoin`

Gmin2:
> Closing this for absence of activity

Hey @vincenzopalazzo i am sorry for my absence it just completely skip my mind, can we push this to complete it !

vincenzopalazzo:
Awesome, thanks!

lampo-cli/src/args.rs
Gmin2: hmm, what alternate do yu suggest ?

vincenzopalazzo:
fell free to reopen this when you will have time!

Issue #245, opened 2024-06-05
Jsonrpc : panic on response


Harshit933:
```
2024-06-05T13:48:55.783Z INFO lampo_jsonrpc Accepting connection: `Ok((UnixStream { fd: FileDesc(OwnedFd { fd: 6 }), local: "/home/doom/.lampo/testnet/lampod.socket" (pathname), peer: (unnamed) }, (unnamed)))`. [lampo-jsonrpc/src/lib.rs:218]
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', /home/doom/Desktop/dev/lampo.rs/lampo-jsonrpc/src/lib.rs:235:68
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Running e3e0b91

vincenzopalazzo:
What is the commit that you are building?

Harshit933:
Latest one

vincenzopalazzo:
Last one tomorrow is not the last one, please add the commit hash reference

Harshit933:
This one e3e0b91

vincenzopalazzo:
I update the issue description with all the details,and marked the comment as off-topics

vincenzopalazzo:
This should be finally fixes when we release the https://github.com/vincenzopalazzo/lampo.rs/pull/246

So for now bumping to the next release

Issue #222, opened 2024-05-16
implementing a public http API


vincenzopalazzo:
I would like to test the flexibility of lampo to see how much time I am able to write an HTTP serever

Issue #182, opened 2024-01-29
bug: Resource temporarily unavailable


Harshit933:
This is the stack trace. When I tried to to ran `lampo-cli getinfo`. 

```
2024-01-29T11:06:27.793Z INFO lampo_channel_manager listening on chain event on the channel manager. [lampod/src/ln/channe_manager.rs:145]
2024-01-29T11:06:27.796Z INFO lampo Litening for in-bound connection on 0.0.0.0:19735. [lampod/src/ln/peer_manager.rs:147]
2024-01-29T11:07:10.151Z INFO jsonrpc blocking with err Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }!. [lampo-jsonrpc/src/lib.rs:225]
2024-01-29T11:07:10.151Z INFO jsonrpc buffer read {"method":"getinfo","params":{},"id":"0". [lampo-jsonrpc/src/lib.rs:232]
```
Although lampo does not crashes but it is not usable. Is this related to #120?

vincenzopalazzo:
>This is the stack trace. 

It is not a stacktrace, is a log line at the info level

>but it is not usable.

define not usable, `WouldBlock` is an expected error while you are trying to write async code

To me this is good, what is the behaviours that you are experiencing?

Harshit933:
I did not get any output after running `lampo-cli getinfo`. This is what I got. 
```
harshit@harshit:~$ lampo-cli getinfo
✗ JSON decode error: Connection reset by peer (os error 104)
```

vincenzopalazzo:
Closing  in favor of https://github.com/vincenzopalazzo/lampo.rs/issues/184

vincenzopalazzo:
I reproduced by changing cloud machine

```
➜  ~ lampo-cli --network testnet fundchannel --node_id 030b686a163aa2bba03cebb8bab7778fac251536498141df0a436d688352d426f6 --amount 4000000000 --public true --node_id 030b686a163aa2bba03cebb8bab7778fac251536498141df0a436d688352d426f6 --addr 65.108.246.14 --port 19735
✗ Error: JSON decode error: Connection reset by peer (os error 104)
➜  ~ lampo-cli --network testnet getinfo
✗ Error: IO error response: No such file or directory (os error 2)
```

I think the bug is inside the lampo-jsonrpc

Harshit933:
[...]

PR #372, opened 2025-03-08
Remove the hand coded unix socket and use the http base API

vincenzopalazzo:
This introduces all the work that is needed to have a basic httpd daemon and remove my hand-coded UNIX socket async server.

This will fix a couple of bugs 

Fixes https://github.com/vincenzopalazzo/lampo.rs/issues/245
Fixes https://github.com/vincenzopalazzo/lampo.rs/issues/243
Fixes https://github.com/vincenzopalazzo/lampo.rs/issues/222
Fixes https://github.com/vincenzopalazzo/lampo.rs/issues/182
Fixes https://github.com/vincenzopalazzo/lampo.rs/issues/139


vincenzopalazzo:
This PR is now including the https://github.com/vincenzopalazzo/lampo.rs/pull/292 PR too because the refactoring was required.

I could do a better job but I am working on this refactoring for too much time, so it is better that we start to ship something!

PR #406, opened 2025-04-05
Remove core_url, core_pass, core_user from Option datatype

Harshit933:
After we removed nakamoto backend, it is better suited to change the datatype from Option<String> to String only

What I have done in this PR?

- Removed the `core_url`, `core_user`, `core_pass` as an option type. 
- Made a new struct with the default configurations if there is no provided inside the cli arguments or `lampo.conf` as suggested by @vincenzopalazzo 
- Made different default cases for different types of network.

### DIFF
--- a/lampo-bdk-wallet/src/lib.rs
+++ b/lampo-bdk-wallet/src/lib.rs
@@ -136,15 +136,9 @@ impl BDKWalletManager {
     }
 
     pub fn build_client(conf: Arc<LampoConf>) -> error::Result<Client> {
-        let url = conf.core_url.as_ref().ok_or(error::anyhow!(
-            "RPC URL is missing from the configuration file"
-        ))?;
-        let user = conf.core_user.as_ref().ok_or(error::anyhow!(
-            "RPC User is missing from the configuration file"
-        ))?;
-        let pass = conf.core_pass.as_ref().ok_or(error::anyhow!(
-            "RPC Password is missing from the configuration file"
-        ))?;
+        let url = &conf.core_url;
+        let user = &conf.core_user;
+        let pass = &conf.core_pass;
         let client = Client::new(url, Auth::UserPass(user.clone(), pass.clone()))?;
         Ok(client)
     }
--- a/lampo-bdk-wallet/src/lib.rs
+++ b/lampo-bdk-wallet/src/lib.rs
@@ -136,9 +136,19 @@ impl BDKWalletManager {
     }
 
     pub fn build_client(conf: Arc<LampoConf>) -> error::Result<Client> {
-        let url = &conf.bitcoind_conf.get_url();
-        let user = &conf.bitcoind_conf.get_user();
-        let pass = &conf.bitcoind_conf.get_pass();
+        let bitcoin_conf = conf
+            .bitcoind_conf
+            .clone()
+            .ok_or(error::anyhow!("Bitcoin conf missing"))?;
+        let url = &bitcoin_conf.url;
+
+        let user = &bitcoin_conf
+            .user
+            .ok_or(error::anyhow!("User not provided"))?;
+        let pass = bitcoin_conf
+            .pass
+            .ok_or(error::anyhow!("Pass not provided"))?;
+
         let client = Client::new(url, Auth::UserPass(user.clone(), pass.clone()))?;
         Ok(client)
     }
--- a/lampo-bdk-wallet/src/lib.rs
+++ b/lampo-bdk-wallet/src/lib.rs
@@ -136,9 +136,9 @@ impl BDKWalletManager {
     }
 
     pub fn build_client(conf: Arc<LampoConf>) -> error::Result<Client> {
-        let url = &conf.core_url;
-        let user = &conf.core_user;
-        let pass = &conf.core_pass;
+        let url = &conf.bitcoind_conf.get_url();
+        let user = &conf.bitcoind_conf.get_user();
+        let pass = &conf.bitcoind_conf.get_pass();
         let client = Client::new(url, Auth::UserPass(user.clone(), pass.clone()))?;
         Ok(client)
     }
--- a/lampo-common/src/btc_conf.rs
+++ b/lampo-common/src/btc_conf.rs
@@ -1,79 +1,43 @@
 use bitcoin::Network;
-use clightningrpc_conf::CLNConf;
+
+use crate::error;
 
 #[derive(Clone, Debug)]
 pub struct BitcoindConf {
-    url: String,
-    user: String,
-    pass: String,
+    pub url: String,
+    pub user: Option<String>,
+    pub pass: Option<String>,
 }
 
-// The default URLs for different bitcoin chains
-pub static BITCOIN_MAINNET_URL: &str = "127.0.0.1:8332";
-pub static BITCOIN_TESTNET_URL: &str = "127.0.0.1:18332";
-pub static BITCOIN_SIGNET_URL: &str = "127.0.0.1:28332";
-pub static BITCOIN_REGTEST_URL: &str = "127.0.0.1:38332";
-pub static BITCOIN_USER: &str = "user";
-pub static BITCOIN_PASS: &str = "pass";
-
 impl BitcoindConf {
-    pub fn get_default_conf(network: Network) -> Self {
+    pub fn get_default_conf(network: Network) -> Result<Self, anyhow::Error> {
         match network {
-            Network::Bitcoin => BitcoindConf {
-                url: BITCOIN_MAINNET_URL.to_string(),
-                user: BITCOIN_USER.to_string(),
-                pass: BITCOIN_PASS.to_string(),
-            },
-            Network::Testnet => BitcoindConf {
-                url: BITCOIN_TESTNET_URL.to_string(),
-                user: BITCOIN_USER.to_string(),
-                pass: BITCOIN_PASS.to_string(),
-            },
-            Network::Signet => BitcoindConf {
-                url: BITCOIN_SIGNET_URL.to_string(),
-                user: BITCOIN_USER.to_string(),
-                pass: BITCOIN_PASS.to_string(),
-            },
-            Network::Regtest => BitcoindConf {
-                url: BITCOIN_REGTEST_URL.to_string(),
-                user: BITCOIN_USER.to_string(),
-                pass: BITCOIN_PASS.to_string(),
-            },
-            Network::Testnet4 => BitcoindConf {
-                url: BITCOIN_TESTNET_URL.to_string(),
-                user: BITCOIN_USER.to_string(),
-                pass: BITCOIN_PASS.to_string(),
-            },
-            // Ideally we can no way enter here, as the network is checked whenever we are
-            // calling this function and if it is not a valid network, the program flow
-            // would not enter here.
-            _ => panic!(
-                "This shouldn't have happened, take a look at your bitcoin network just in case"
-            ),
+            Network::Bitcoin => Ok(BitcoindConf {
+                url: "127.0.0.1:8332".to_string(),
+                user: None,
+                pass: None,
+            }),
+            Network::Testnet => Ok(BitcoindConf {
+                url: "127.0.0.1:18332".to_string(),
+                user: None,
+                pass: None,
+            }),
+            Network::Signet => Ok(BitcoindConf {
+                url: "127.0.0.1:28332".to_string(),
+                user: None,
+                pass: None,
+            }),
+            Network::Regtest => Ok(BitcoindConf {
+                url: "127.0.0.1:38332".to_string(),
+                user: None,
+                pass: None,
+            }),
+            Network::Testnet4 => Ok(BitcoindConf {
+                url: "127.0.0.1:18332".to_string(),
+                user: None,
+                pass: None,
+            }),
+            _ => error::bail!("Network not supported"),
         }
-    }
-
-    pub fn get_url(&self) -> String {
-        self.url.clone()
-    }
-
-    pub fn get_user(&self) -> String {
-        self.user.clone()
-    }
-
-    pub fn get_pass(&self) -> String {
-        self.pass.clone()
-    }
-
-    pub fn set_url(&mut self, url: String) {
-        self.url = url
-    }
-
-    pub fn set_user(&mut self, user: String) {
-        self.user = user
-    }
-
-    pub fn set_pass(&mut self, pass: String) {
-        self.pass = pass
     }
 }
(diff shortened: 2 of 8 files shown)


lampo-bdk-wallet/src/lib.rs
```diff
@@ -136,15 +136,9 @@ impl BDKWalletManager {
...
     pub fn build_client(conf: Arc<LampoConf>) -> error::Result<Client> {
-        let url = conf.core_url.as_ref().ok_or(error::anyhow!(
-            "RPC URL is missing from the configuration file"
-        ))?;
-        let user = conf.core_user.as_ref().ok_or(error::anyhow!(
-            "RPC User is missing from the configuration file"
-        ))?;
-        let pass = conf.core_pass.as_ref().ok_or(error::anyhow!(
-            "RPC Password is missing from the configuration file"
-        ))?;
```
vincenzopalazzo, line 147: If you want really remove these check, that are duplicated, as you correctly noted

a good API design for this could be `conf.bitcoind_conf()` and return a struct 

```rust
struct BitcoindConf {
   user: String,
   pass: String,
   url: String
}
```

and then for every network choose the default values!

lampo-common/src/conf.rs
```diff
@@ -46,9 +46,9 @@ impl Default for LampoConf {
             port: 19735,
             root_path: lampo_home,
             node: "core".to_owned(),
-            core_url: None,
-            core_user: None,
-            core_pass: None,
+            core_url: "".to_string(),
+            core_user: "".to_string(),
+            core_pass: "".to_string(),
```
vincenzopalazzo, line 51: You are losing the Rust type system propriety!

Now you need to check for `core_url.is_empty()` that has the same semantic meaning of `Option<String>` but you are losing the compiler check!

lampo-common/src/conf.rs
```diff
@@ -180,26 +180,26 @@ impl TryFrom<String> for LampoConf {
...
-                .get_conf("core-url")
-                .map_err(|err| anyhow::anyhow!("{err}"))?;
-            // If the value isn't none, strip the value of whitespace
-            core_url = core_url.map(|url| url.to_trimmed());
+        let Some(core_url) = conf
+            .get_conf("core-url")
+            .map_err(|err| anyhow::anyhow!("{err}"))?
+        else {
+            anyhow::bail!("Core URL must be specified")
+        };
```
vincenzopalazzo, line 188: This look wired, you are already returning an error with `map_err()`? why you are not writing it in this way

```suggestion
        let Some(core_url) = conf
            .get_conf("core-url")
            // if there is a error working with the config file internal
            .map_err(|err| anyhow::anyhow!("{err}"))?
            // if there is no value inside the configuration file
            .ok_or("Core URL must be specified")?;
```

it is much more easy to read IMHO and understand what it is going on

lampo-common/src/conf.rs
```diff
@@ -180,26 +180,26 @@ impl TryFrom<String> for LampoConf {
...
-                .get_conf("core-pass")
-                .map_err(|err| anyhow::anyhow!("{err}"))?;
-            core_pass = core_pass.map(|pass| pass.to_trimmed());
-        }
+        let Some(core_pass) = conf
+            .get_conf("core-pass")
+            .map_err(|err| anyhow::anyhow!("{err}"))?
+        else {
+            anyhow::bail!("Core Password must be specified")
+        };
```
vincenzopalazzo, line 202: same here!

lampod-cli/src/args.rs
```diff
@@ -79,13 +79,13 @@ impl TryInto<LampoConf> for LampoCliArgs {
             conf.node = node.clone();
         }
         if self.bitcoind_url.is_some() {
-            conf.core_url = self.bitcoind_url;
+            conf.core_url = self.bitcoind_url.unwrap();
         }
```
vincenzopalazzo, line 83: `.unwrap()` cause program panic! again and as in all my review you should avoid to use it:)

it is so clean write it in the following way

```rust
if let Some(bitcoin_url) = self.bitcoin_url {
 conf.core_url = bitcoin_url;
}
```

vincenzopalazzo (changes requested):
Reading the code you are making them optional anyway, but you are not using the rust type system

The only change that this PR need to address, is when the value is none inside the `conf.rs` file, we should choose better default value! Other part of the codebase will be not effected.

lampo-bdk-wallet/src/lib.rs
Harshit933: This was a nice design, thanks

lampo-common/src/conf.rs
Harshit933: Noted, thanks

Harshit933:
@vincenzopalazzo There could be one thing that I think I can do more to this PR. For now I have put default values for user and pass for all the cases of network to be `user` and `pass` respectively (you can see [here](https://github.com/vincenzopalazzo/lampo.rs/pull/406/files#diff-5b77d4a63e32976e70a47021361a1024a5869122f1cbe2855e2d005d9f049964R16). But, I don't think so this is a good thing to do? Maybe we should just put default `url` for all the different type of networks, like for mainnet the url would be `127.0.0.1:8332`. And if there is no user and pass specified we can just throw an error and not try to put the default values.

lampo-common/src/btc_conf.rs
```diff
@@ -0,0 +1,79 @@
...
+    pass: String,
+}
+
+// The default URLs for different bitcoin chains
+pub static BITCOIN_MAINNET_URL: &str = "127.0.0.1:8332";
+pub static BITCOIN_TESTNET_URL: &str = "127.0.0.1:18332";
+pub static BITCOIN_SIGNET_URL: &str = "127.0.0.1:28332";
+pub static BITCOIN_REGTEST_URL: &str = "127.0.0.1:38332";
+pub static BITCOIN_USER: &str = "user";
+pub static BITCOIN_PASS: &str = "pass";
```
vincenzopalazzo, line 17: Check if there is a default user and pass inside the Bitcoin Core docs; otherwise make them optional

lampo-common/src/btc_conf.rs
```diff
@@ -0,0 +1,79 @@
...
+    url: String,
+    user: String,
+    pass: String,
+}
+
+// The default URLs for different bitcoin chains
+pub static BITCOIN_MAINNET_URL: &str = "127.0.0.1:8332";
+pub static BITCOIN_TESTNET_URL: &str = "127.0.0.1:18332";
+pub static BITCOIN_SIGNET_URL: &str = "127.0.0.1:28332";
+pub static BITCOIN_REGTEST_URL: &str = "127.0.0.1:38332";
```
vincenzopalazzo, line 15: What is the meaning of having a bitcoin configuration if you have a global variable? In this paradigm that we are using, this does not make any sense because you are just adding the value inside the `get_default_conf `

So put the value directly in there

lampo-common/src/btc_conf.rs
```diff
@@ -0,0 +1,79 @@
...
+            },
+            Network::Testnet4 => BitcoindConf {
+                url: BITCOIN_TESTNET_URL.to_string(),
+                user: BITCOIN_USER.to_string(),
+                pass: BITCOIN_PASS.to_string(),
+            },
+            // Ideally we can no way enter here, as the network is checked whenever we are
+            // calling this function and if it is not a valid network, the program flow
+            // would not enter here.
+            _ => panic!(
```
vincenzopalazzo, line 50: No, this can happen for this reason, the network in rust-bitcoin is not exhaustive (again the type system is already telling you this)

You can look it inside the documentation https://docs.rs/bitcoin/latest/bitcoin/enum.Network.html

>Non-exhaustive enums could have additional variants added in future. Therefore, when matching against variants of non-exhaustive enums, an extra wildcard arm must be added to account for any future variants.

So this will panic at some point in life, so it is better return an error and remove the commit

lampo-common/src/btc_conf.rs
```diff
@@ -0,0 +1,79 @@
...
+        self.url = url
+    }
+
+    pub fn set_user(&mut self, user: String) {
+        self.user = user
+    }
+
+    pub fn set_pass(&mut self, pass: String) {
+        self.pass = pass
+    }
```
vincenzopalazzo, line 78: We are not in java :) Make the field public directly! and remove all these get and set

lampo-common/src/conf.rs
```diff
@@ -6,6 +6,7 @@ use clightningrpc_conf::{CLNConf, SyncCLNConf};
 pub use bitcoin::Network;
 pub use lightning::util::config::UserConfig;
 
+pub use crate::btc_conf::BitcoindConf;
```
vincenzopalazzo, line 9: please add a space
```suggestion
pub use crate::btc_conf::BitcoindConf;

```

lampo-common/src/conf.rs
```diff
@@ -58,6 +54,8 @@ impl Default for LampoConf {
             api_host: "127.0.0.1".to_owned(),
             api_port: 7878,
             reindex: None,
+            // By default network is testnet as mentioned above
+            bitcoind_conf: BitcoindConf::get_default_conf(Network::Testnet),
```
vincenzopalazzo, line 58: Keep this None, and then set inside the `TryFrom` after you are setting up the network

lampo-common/src/conf.rs
```diff
@@ -179,27 +177,32 @@ impl TryFrom<String> for LampoConf {
...
 
-            core_user = conf
-                .get_conf("core-user")
-                .map_err(|err| anyhow::anyhow!("{err}"))?;
-            core_user = core_user.map(|user| user.to_trimmed());
+        // Override if there is some config specified!
+        if let Some(bitcoin_url) = conf
+            .get_conf("core-url")
+            .map_err(|err| anyhow::anyhow!("{err}"))?
+        {
```
vincenzopalazzo, line 189: split 
```suggestion
        let bitcoin_url = conf
            .get_conf("core-url")
            .map_err(|err| anyhow::anyhow!("{err}"))?;
            
        if let Some(bitcoin_url) = bitcoin_url {
```

is much readble in this way

lampo-common/src/conf.rs
```diff
@@ -179,27 +177,32 @@ impl TryFrom<String> for LampoConf {
...
+            bitcoind_conf.set_pass(bitcoin_pass);
+        };
+
+        if let Some(bitcoin_user) = conf
+            .get_conf("core-user")
+            .map_err(|err| anyhow::anyhow!("{err}"))?
+        {
+            bitcoind_conf.set_user(bitcoin_user);
+        };
```
vincenzopalazzo, line 205: Same for these values!

lampo-testing/src/lib.rs
```diff
@@ -105,9 +105,9 @@ impl LampoTesting {
         lampo_conf.api_port = port::random_free_port().unwrap().into();
         log::info!("listening on port `{}`", lampo_conf.api_port);
         let core_url: String = format!("http://127.0.0.1:{}", btc.port);
-        lampo_conf.core_pass = Some(btc.pass.clone());
-        lampo_conf.core_url = Some(core_url);
-        lampo_conf.core_user = Some(btc.user.clone());
+        lampo_conf.bitcoind_conf.set_pass(btc.pass.clone());
+        lampo_conf.bitcoind_conf.set_url(core_url);
+        lampo_conf.bitcoind_conf.set_user(btc.user.clone());
```
vincenzopalazzo, line 110: nit: no need to use the set in rust!

vincenzopalazzo (changes requested):
Ok we are getting there, now you can see how the PR is much much smaller and much readble

lampo-common/src/btc_conf.rs
Harshit933: Maybe juggling between languages is not that great hahahah

lampo-common/src/btc_conf.rs
Harshit933: I didn't find any, moving this to optional

lampo-common/src/btc_conf.rs
Harshit933: Got it, thanks

lampo-common/src/btc_conf.rs
Harshit933: The network here is formed from the this function

```rust
fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s {
            "bitcoin" => Ok(Network::Bitcoin),
            // For user-side compatibility, testnet3 is retained as testnet
            "testnet" => Ok(Network::Testnet),
            "testnet4" => Ok(Network::Testnet4),
            "signet" => Ok(Network::Signet),
            "regtest" => Ok(Network::Regtest),
            _ => Err(ParseNetworkError(s.to_owned())),
        }
    }
```
Thus, for now there could be only four possible networks, otherwise we return the error before going to this condition.

The documentation says that in future, there could be addition of other networks so the match here is marked as non exhaustive.

So, I don't think so, this condition would ever hit for now.

Harshit933: Returning an error from here, I got it what you're saying.

lampo-common/src/conf.rs
Harshit933: Done

vincenzopalazzo:
Can you rebase the PR on the current main? thanks

vincenzopalazzo:
@Harshit933 are you willing to put this PR on the finish line?

vincenzopalazzo (commented):
**Production Readiness: NO-GO.** The configuration migration is incomplete and the new defaults select incorrect RPC ports.

### Positive Aspects

- Centralizing Bitcoin Core settings in BitcoindConf is the right direction.

### Issues & Suggestions

- **Critical:** lampo-common/src/btc_conf.rs:7-9 keeps user and pass as optional strings, while call sites still fail at runtime when either value is absent. This does not complete the PR's stated migration. Make the fields concrete strings or narrow the PR's contract and preserve explicit validation consistently.
- **Critical:** lampo-common/src/btc_conf.rs:16-39 assigns the wrong RPC ports for Signet, Regtest, and Testnet4. They should be 38332, 18443, and 48332; the current values can connect to the wrong service or fail to start.
- **Medium:** LampoConf::default() leaves bitcoind_conf unset, but get_bitcoind() unwraps it and the CLI conversion requires it to be present. Initialize the default configuration or return a fallible/optional value instead of introducing a default-path panic/error.

### Summary

Please finish the type migration, correct the network defaults, and add coverage for every Bitcoin network plus the no-config startup path.

Panel: 3 independent models (consensus 2/3 to flag). Targeted checks on the PR head: static source verification only; no CI checks are published for this head.

PR #413, opened 2025-04-11
lnprototest: be able to run lnprototest again

vincenzopalazzo:
This is fixing lampo in order to run lnprototest and allow @Psycho-Pirate to work an the runner and the extension of the lampo tests, itself


tests/lnprototest/dist/lampo_lnprototest-0.0.1-py3-none-any.whl
vincenzopalazzo, line 1: these files need to be removed from the git history

PR #414, opened 2025-04-11
Create a wallet using custom private key and channel secrets

Psycho-Pirate:
This PR allows Lampo to use custom private key and channel secrets mentioned in the config file.

### DIFF
--- a/lampo-bdk-wallet/src/lib.rs
+++ b/lampo-bdk-wallet/src/lib.rs
@@ -95,7 +95,7 @@ impl BDKWalletManager {
         conf: Arc<LampoConf>,
         xprv: PrivateKey,
         channel_keys: Option<String>,
-    ) -> error::Result<(PersistedWallet<Connection>, Arc<Connection>, LampoKeys)> {
+    ) -> error::Result<(PersistedWallet<Connection>, Connection, LampoKeys)> {
         let ldk_keys = if channel_keys.is_some() {
             LampoKeys::with_channel_keys(xprv.inner.secret_bytes(), channel_keys.unwrap())
         } else {
@@ -132,7 +132,7 @@ impl BDKWalletManager {
 
         let descriptor = wallet.public_descriptor(KeychainKind::Internal);
         log::info!("descriptor: {descriptor}");
-        Ok((wallet, Arc::new(db), ldk_keys))
+        Ok((wallet, db, ldk_keys))
     }
 
     pub fn build_client(conf: Arc<LampoConf>) -> error::Result<Client> {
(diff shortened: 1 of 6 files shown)


lampo-bdk-wallet/src/lib.rs
```diff
@@ -95,7 +95,7 @@ impl BDKWalletManager {
         conf: Arc<LampoConf>,
         xprv: PrivateKey,
         channel_keys: Option<String>,
-    ) -> error::Result<(PersistedWallet<Connection>, Arc<Connection>, LampoKeys)> {
+    ) -> error::Result<(PersistedWallet<Connection>, Connection, LampoKeys)> {
```
vincenzopalazzo, line 98: why you are removing `Arc` here?

lampo-common/src/wallet.rs
```diff
@@ -24,6 +24,11 @@ pub trait WalletManager: Send + Sync {
     where
         Self: Sized;
 
+    /// generate a new wallet from a private key and channel secrets 
+    async fn create_using_private_key(conf: Arc<LampoConf>, xprv: PrivateKey, channel_keys: Option<String> )  -> error::Result<Self>
```
vincenzopalazzo, line 28: we can mark this unsafe somehow?

lampod-cli/src/main.rs
```diff
@@ -91,68 +94,97 @@ async fn run(args: LampoCliArgs) -> error::Result<()> {
...
-    // 2. lampo is running on os where there is a wallet, lampo will load the seeds from wallet:
-    //   2.1: The user keep specify --restore-wallet, so lampo should return an error an tell the user that there is already a wallet
-    //   2.2: The user do not specify the --restore-wallet, so lampo load from disk the file, if there is no file it return an error
-    // FIXME: there is a problem of code duplication here, we should move this code in utils functions.
-    let wallet = if restore_wallet {
-        if Path::new(&format!("{}/wallet.dat", words_path)).exists() {
-            // Load the mnemonic from the file
-            let mnemonic = load_words_from_file(format!("{}/wallet.dat", words_path))?;
-            let wallet = match client.kind() {
+    let wallet = match (lampo_conf.private_key.clone(), lampo_conf.channels_keys.clone()) {
```
vincenzopalazzo, line 97: Wow this match is really complex!

No you need to e explain me this sorry :) I am not able to simply understand while reading

lampod-cli/src/main.rs
```diff
@@ -91,68 +94,97 @@ async fn run(args: LampoCliArgs) -> error::Result<()> {
...
+        (Some(private_key), Some(channels_keys)) => {
+            // Create wallet with both key and secrets
+            let secret_key = SecretKey::from_str(&private_key);
+            let private_key = PrivateKey {
+                compressed: true,
+                network: NetworkKind::from(lampo_conf.network.clone()), // change to Mainnet/Testnet if needed
```
vincenzopalazzo, line 103: We should never do it with mainer and testnet IMHO just for regtest

vincenzopalazzo (commented):
I think this change is too big for a simple operation like this one! Mh I need to rework this code!

PR #420, opened 2025-04-14
BAD MAC error fixed

Psycho-Pirate:
This commit fixes the BAD MAC error while running Lnprototest


vincenzopalazzo:
Including only the fix in here https://github.com/vincenzopalazzo/lampo.rs/pull/419 the other is an hack that it is solved by the linked PR

Thanks for catching it

PR #435, opened 2025-05-10
feat: added clap for `lampo-cli`

Gmin2:
Do we need to introduce clap for `lampod-cli` also? as it would be more consistent

### DIFF
--- a/lampo-cli/src/args.rs
+++ b/lampo-cli/src/args.rs
@@ -1,138 +1,66 @@
 use std::collections::HashMap;
 
-use radicle_term as term;
+use clap::Parser;
 
 use lampo_common::error;
 use lampo_common::json;
 
-#[derive(Debug)]
+#[derive(Debug, Parser)]
+#[command(name = "lampod-cli")]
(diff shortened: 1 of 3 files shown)


lampo-cli/src/args.rs
```diff
@@ -1,138 +1,66 @@
...
 
 use lampo_common::error;
 use lampo_common::json;
 
-#[derive(Debug)]
+#[derive(Debug, Parser)]
+#[command(name = "lampod-cli")]
+#[command(about = "Lampo Daemon command line", version = env!("CARGO_PKG_VERSION"))]
+#[command(before_help = "Common `lampod-cli` commands used to interact with the lampo daemon")]
+#[command(after_help = "Example: lampod-cli get_block height=123 verbose=true")]
```
vincenzopalazzo, line 12:
```suggestion
```

vincenzopalazzo (commented):
Mh I think you started with the wrong one. This is a little bit tricky to do and doing the 1-1 map of the previous solution is not what we are looking for.

Probably you can do something similar to what you propose in `lampod-cli`, but in another PR

PR #441, opened 2025-05-23
bdk: skip the sync network from genesis

vincenzopalazzo:
When a wallet start fresh we can skip to sync
the network from the genesis block to the current
block height.

Suggested-by: @ValuedMammal

### DIFF
--- a/lampo-bdk-wallet/src/lib.rs
+++ b/lampo-bdk-wallet/src/lib.rs
@@ -4,6 +4,7 @@ use std::sync::Arc;
 
 use bdk_bitcoind_rpc::bitcoincore_rpc::{Auth, Client, RpcApi};
 use bdk_bitcoind_rpc::Emitter;
+use bdk_wallet::chain::BlockId;
 use bdk_wallet::keys::bip39::Mnemonic;
 use bdk_wallet::keys::bip39::{Language, WordCount};
 use bdk_wallet::keys::{DerivableKey, ExtendedKey, GeneratableKey, GeneratedKey};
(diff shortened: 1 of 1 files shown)


ValuedMammal:
Hey @vincenzopalazzo I haven't tested it in `lampo` but I would do something like this to ensure that syncing starts from either the latest wallet tip or the `reindex_from` height, whichever is greater.

```rust
    let rpc_client = self.rpc.clone();
    let mut wallet = self.wallet.lock().await;
    let mut wallet_tip = wallet.latest_checkpoint();
    let mut start_height = wallet_tip.height();

    // This should tell the wallet, hey from now on you will have to reindex from this height.
    if let Some(height) = self.reindex_from {
        let height = height.to_consensus_u32();
        if height > wallet_tip.height() {
            // Insert a checkpoint into the wallet to avoid scanning the entire chain.
            let hash = rpc_client.get_block_hash(height as u64)?;
            let block = BlockId { height, hash };
            let new_tip = wallet_tip.insert(block);
            let update = bdk_wallet::Update {
                chain: Some(new_tip),
                ..Default::default()
            };
            wallet.apply_update(update)?;
            wallet_tip = wallet.latest_checkpoint();
            start_height = wallet_tip.height();
        }
    }

    drop(wallet); // Drop the mutex

    tokio::spawn(async move {
        let mut emitter = Emitter::new(rpc_client.as_ref(), wallet_tip, start_height);
        
        while let Some(emission) = emitter.next_block()? {
            sender.send(Emission::Block(emission))?;
        }
        sender.send(Emission::Mempool(emitter.mempool()?))?;
        
        Ok::<_, error::Error>(())
    });
```

ValuedMammal:
Ok I see what you mean now, that makes sense.

vincenzopalazzo:
>I assumed the user would always configure a reindex_from height in order to benefit from the skip-ahead logic.

That makes sense, and maybe I'm trying to be smarter here and assuming too much. However, in my experience, the user will keep using the defaults

ValuedMammal:
Yes it looks like it will help to skip syncing from genesis, but I may have misunderstood the purpose of `reindex_from`. If it is supposed to be used to rescan from an _earlier_ height (as the name implies), then I imagine you would find the nearest wallet checkpoint at or below the reindex height and use that when constructing the Emitter.. but maybe that's out of scope here.

As for the CI failure, I don't know.

PR #452, opened 2025-06-06
Macros/custom txs

vincenzopalazzo:


vincenzopalazzo:
This PR took too much popularity, and it is time that it dies with the PoC; now there are better PoC out there

PR #513, opened 2026-03-11
fix: handle reindex_from when height is below wallet tip

### LINKED ISSUE #444: bug when reindexing a chain from a previous height
Yes it looks like it will help to skip syncing from genesis, but I may have misunderstood the purpose of `reindex_from`. If it is supposed to be used to rescan from an _earlier_ height (as the name implies), then I imagine you would find the nearest wallet checkpoint at or below the reindex height and use that when constructing the Emitter.. but maybe that's out of scope here.

As for the CI failure, I don't know.

_Originally posted by @ValuedMammal in https://github.com/vincenzopalazzo/lampo.rs/issues/441#issuecomment-2920102944_

vincenzopalazzo:
## Summary
- When `reindex_from` is set to a height **lower** than the current wallet tip, the wallet now finds the nearest checkpoint at or below the requested height and rolls back to it, so the `Emitter` rescans blocks from that point
- Previously this case was silently ignored — only skip-ahead (`height > tip`) worked, making backward rescans ineffective

## Test plan
- [ ] Start a wallet with existing chain state at height N
- [ ] Set `reindex_from` to a height < N
- [ ] Verify the wallet rescans from the nearest checkpoint at or below the requested height
- [ ] Verify the existing skip-ahead behavior (`reindex_from` > tip) still works
- [ ] Verify fresh wallet behavior (tip at 0) still auto-detects chain height

Closes #444


vincenzopalazzo:
## Review vs main + ldk-node

**The bug is real** (issue #444). On `main`, `reindex_from` only skip-aheads:

```
if height > wallet_tip.height() { insert checkpoint }
```

A value *below* the current tip was a silent no-op, so a restored wallet could not rescan missed history.

**ldk-node** has no `reindex` knob; the BDK wallet rides `synchronize_listeners`. This is a lampo operator tool.

**What landed in this rebase (on current main):**
- Keep skip-ahead **and persist it** (already on main).
- Add rollback to the nearest checkpoint `<= height`.
- Persist the rollback + `set_wallet_scan_height` (the original PR applied in memory only).
- Error if no checkpoint exists that far back (no silent `unwrap_or(tip)`).

`cargo test -p lampo-bdk-wallet --lib` green.

PR #515, opened 2026-03-11
fix: add fee estimation fallback for IBD and insufficient data

### LINKED ISSUE #244: `Fees` command returns invalid response
```
[doom@archlinux ~]$ lampo-cli --network testnet fees
{
  "non_anchor_channel": null,
  "channel_close_minimum": null,
  "min_allowed_anchor_channel_remote": null,
  "anchor_chanenl": null,
  "on_chain_sweep": null,
  "min_allowed_non_anchor_channel_remote": null
}
```

vincenzopalazzo:
## Summary

Fixes #244 — `fees` command returns all null values when Bitcoin Core is in Initial Block Download (IBD) or has insufficient fee estimation data.

**Three problems fixed:**

- **`fee_rate_estimation()` in `lampo-chain`**: When `estimatesmartfee` fails (IBD, insufficient data), now falls back to `minimum_mempool_fee()` instead of returning an error that gets silently swallowed
- **`get_est_sat_per_1000_weight()` in `lampod`**: Was hardcoded to return `256` — now actually queries the backend with proper confirmation targets (1/6/12/100 blocks depending on urgency)
- **`estimated_fees()` response**: Includes a `"warning"` field when fallback values are used, so users understand why estimates may be less accurate

A floor constant `FEERATE_FLOOR_SATS_PER_KW = 253` (matching LDK's internal floor) ensures we never return zero fee rates, which would block channel operations.

## Test plan

- [ ] Verify `lampo-cli fees` returns numeric values instead of nulls when the backend node is syncing
- [ ] Verify the `warning` field appears in the response when fallback values are used
- [ ] Verify normal fee estimation still works correctly when the node is fully synced
- [ ] Verify regtest still returns hardcoded 256

### DIFF
--- a/lampo-chain/src/lib.rs
+++ b/lampo-chain/src/lib.rs
@@ -148,14 +148,20 @@ impl Backend for LampoChainSync {
             .call_method::<json::Value>("estimatesmartfee", &[blocks.into()])
             .await?;
         let resp: FeeRate = json::from_value(resp)?;
-        if let Some(errs) = resp.errors {
-            return Err(error::anyhow!("Error in fee rate estimation: {:?}", errs).into());
+        if resp.errors.is_some() || resp.feerate.is_none() {
+            // When estimatesmartfee fails (e.g. during Initial Block Download
+            // or when the node has insufficient data), fall back to the
+            // minimum mempool fee. This prevents returning null fee estimates
+            // to the caller. See: https://github.com/vincenzopalazzo/lampo.rs/issues/244
+            log::warn!(
+                "estimatesmartfee failed for {} blocks (errors: {:?}), falling back to minimum mempool fee",
+                blocks,
+                resp.errors
+            );
+            return self.minimum_mempool_fee().await;
         }
-        let Some(feerate) = resp.feerate else {
-            return Err(error::anyhow!("No fee rate estimation available").into());
-        };
         // estimate fee rate in BTC/kvB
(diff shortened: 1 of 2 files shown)


vincenzopalazzo:
## Review vs main + ldk-node — do not merge

**Mostly superseded, and the remaining patch would regress.**

On current `main`:
- `get_est_sat_per_1000_weight` already reads the **fee cache** (not hardcoded 256).
- Regtest returns **250 sat/kW** (ldk-node parity).
- 253 floor / I04 funder is already handled.

This PR would replace the cache with `block_on(fee_rate_estimation)` on **every LDK tick**. ldk-node never does that: it refreshes a cache in the background and falls back to 1 sat/vB (250 sat/kW) on regtest/signet, or skips on testnet.

**Still true:** mainnet IBD `estimatesmartfee` errors are not mapped to mempool-min inside `fee_rate_estimation_with_mode`. That is a 10-line follow-up in the chain crate only — keep the cache, do not change `FeeEstimator`.

Closing this PR. Open a tiny IBD-only patch if we still see null `fees` on mainnet IBD.

vincenzopalazzo:
Superseded by the fee cache + 253 floor + regtest 250 sat/kW already on main. A cache-bypass block_on on every LDK tick would regress vs ldk-node. IBD mempool-min fallback can be a 10-line follow-up in fee_rate_estimation_with_mode only.

Issue #545, opened 2026-06-03
Design: unified chain sync (align with ldk-node, fix dual RPC pipelines)


vincenzopalazzo:
## Summary

Lampo runs **two competing chain-sync pipelines** on the same bitcoind RPC connection:

1. **LDK** `synchronize_listeners` (channel manager + chain monitor only) — started from persisted height ~190k
2. **BDK** `Emitter` full `getblock` scan (on-chain wallet) — started from checkpoint ~21k

On production Signet (Ocean, ~307k blocks), this causes multi-day catch-up, RPC saturation, and **LDK listener sync that never logs completion** while BDK `getblock` dominates logs. `getinfo` shows chain tip but `wallet_height` stays at 21k for days.

**ldk-node** (used by **ldk-server**) avoids this by registering the on-chain BDK wallet as an LDK `Listen` target and running **one** `synchronize_listeners` pass, then a unified poll loop — with retries, completion logs, and sync timestamps in `GetNodeInfo`.

## Design document

Full design (problem, comparison, phased solution, key decisions, PR plan):

**[`docs/designs/unified-chain-sync.md`](https://github.com/vincenzopalazzo/lampo.rs/blob/main/docs/designs/unified-chain-sync.md)**

> **Note:** The design file is added in the local tree; merge/commit the doc path above or paste from PR that adds `docs/designs/unified-chain-sync.md`.

## Proposed direction (high level)

| Phase | What |
|-------|------|
| **A** | `ChainSyncCoordinator` + gate BDK Emitter until LDK listeners synced; extend `getinfo` with sync progress; fix startup order |
| **B** | Implement `Listen` on BDK wallet; include wallet in `synchronize_listeners` (ldk-node parity); deprecate parallel Emitter for initial catch-up |
| **C** | `fast_sync` / `sync_mode` config ([#540](https://github.com/vincenzopalazzo/lampo.rs/issues/540)); `sync_wallets` JSON-RPC |

## Production evidence (Jun 2026)

- ~11h runtime: wallet RPC scan 21k → ~91k (~24%); **no** `Chain listeners synced` log
- ~140k `getblock` calls; 0 ERROR/WARN
- Related: [#540](https://github.com/vincenzopalazzo/lampo.rs/issues/540) (fast sync), [#444](https://github.com/vincenzopalazzo/lampo.rs/issues/444) (`reindex` semantics)

## PR plan (from design)

1. Coordinator + `getinfo` progress fields
[...]

PR #546, opened 2026-06-03
sync: add ChainSyncCoordinator and getinfo progress fields

vincenzopalazzo:
First PR of the unified chain sync design ([#545](https://github.com/vincenzopalazzo/lampo.rs/issues/545)), using the revised **decoupled** approach: keep the unified sync win without coupling the on-chain wallet to LDK.

## Summary

- Add a backend-agnostic `ChainSyncCoordinator` in `lampo-common` (`chainsync.rs`). Pure state — atomics + a `tokio::sync::watch` channel, **no LDK and no BDK types**. Tracks the initial-sync lifecycle (`PendingInitialSync -> ListenersSynced -> Running`) and the live wallet scan height.
- Extend `getinfo` (`GetInfo`) with four progress fields: `wallet_scan_height`, `chain_listeners_synced`, `initial_sync_complete`, `sync_in_progress`.
- Wire one coordinator into `LampoDaemon` (`chain_sync()` accessor) and populate the `getinfo` fields from it.
- Also reworks `docs/designs/unified-chain-sync.md` to the decoupled design: confine the LDK<->wallet coupling to a `Listen` adapter inside `lampo-chain` instead of `impl Listen` on the BDK wallet, express unified sync as a backend-agnostic coordinator, and add a Phase D sketch where an Esplora/Electrum backend needs zero wallet changes.

## Scope / non-goals

- The coordinator is wired in but **not driven yet** — it stays in `PendingInitialSync`, so the new fields report `false`/`true`/`None` defaults for now. Later PRs (gate the BDK `Emitter`, lampo-native wallet `apply_block`, the `Listen` adapter) call `mark_listeners_synced`/`mark_running`, after which `getinfo` reflects real progress with no further changes here.
- No behavior change beyond the additive `getinfo` fields. No Phase B/C work.

## Test plan

- `cargo test` green; 4 new unit tests cover coordinator state transitions (initial state, forward-only `mark_listeners_synced`, no-regress from `Running`, scan-height roundtrip).
- `cargo check --workspace --all-targets` clean (incl. `lampo-testing` and `tests/` which deserialize `GetInfo`).
- `cargo fmt --all -- --check` clean.

### DIFF
--- a/docs/designs/unified-chain-sync.md
+++ b/docs/designs/unified-chain-sync.md
@@ -1,17 +1,39 @@
-# Unified Chain Sync: Align Lampo with ldk-node
+# Unified Chain Sync: Align Lampo with ldk-node *without losing backend/wallet pluggability*
 
 | Field | Value |
 |-------|--------|
-| Status | Draft |
-| Author | Vincenzo Palazzo (with Grok-assisted analysis) |
+| Status | Draft (rev. 2 — decoupled) |
+| Author | Vincenzo Palazzo (with Grok-assisted analysis; rev. 2 decoupling pass) |
 | Date | 2026-06-03 |
-| Related issues | [#540](https://github.com/vincenzopalazzo/lampo.rs/issues/540) (fast sync), [#444](https://github.com/vincenzopalazzo/lampo.rs/issues/444) (`reindex` semantics) |
+| Related issues | [#545](https://github.com/vincenzopalazzo/lampo.rs/issues/545), [#540](https://github.com/vincenzopalazzo/lampo.rs/issues/540) (fast sync), [#444](https://github.com/vincenzopalazzo/lampo.rs/issues/444) (`reindex` semantics) |
 
 ## Summary
 
-Lampo currently runs **two independent, competing chain-sync pipelines** over the same Bitcoin Core RPC backend: LDK `synchronize_listeners` (channel manager + chain monitor) and a separate BDK `Emitter` full-block scan (on-chain wallet). Production on Ocean Signet (~307k blocks) shows multi-day catch-up, RPC saturation, and LDK listener sync that never logs completion while the BDK path dominates `getblock` traffic.
+Lampo currently runs **two independent, competing chain-sync pipelines** — over **two separate Bitcoin Core RPC clients**: LDK `synchronize_listeners` (channel manager + chain monitor, via `lightning-block-sync`'s `RpcClient`) and a separate BDK `Emitter` full-block scan (on-chain wallet, via `bdk_bitcoind_rpc::bitcoincore_rpc::Client`). Production on Ocean Signet (~307k blocks) shows multi-day catch-up, RPC saturation, and LDK listener sync that never logs completion while the BDK path dominates `getblock` traffic.
 
-**ldk-node** (used by **ldk-server**) solves this by registering the on-chain BDK wallet as an LDK `Listen` target and driving **one** `synchronize_listeners` pass, then a unified poll loop. This document proposes converging Lampo toward that model while preserving Lampo’s modular crates and optional checkpoint/fast-sync for signet/dev.
+**ldk-node** (used by **ldk-server**) solves the efficiency problem by registering the on-chain BDK wallet as an LDK `Listen` target and driving **one** `synchronize_listeners` pass, then a unified poll loop.
+
+This document proposes converging Lampo toward that **outcome** (one RPC stream, observable completion, no duplicate scan) **while explicitly preserving Lampo's modular design**, where the Lightning side and the Bitcoin side live behind traits and can be replaced independently. Concretely we **diverge from ldk-node in one place**: we do **not** implement LDK's `Listen` on the wallet. Instead the unavoidable LDK↔wallet coupling is **confined to the bitcoind backend crate (`lampo-chain`)** via a thin adapter, and unified sync is expressed as a **backend-agnostic** responsibility coordinated through `lampo-common`. This keeps the chain `Backend` swappable (Esplora/Electrum next) and the on-chain wallet swappable, with neither side leaking into the other.
+
+## Design principle: backend-confined coupling (the three seams)
+
+Lampo's replaceability rests on three trait seams. The whole point of this design is to keep them intact.
+
+| Seam | Trait / type | Today's impl | What it abstracts |
+|------|--------------|--------------|-------------------|
+| Chain source | `Backend` (`lampo-common/src/backend.rs`) | `LampoChainSync` (`lampo-chain`) | How blocks/txs are fetched and how consumers are driven to tip |
+| On-chain wallet | `WalletManager` (`lampo-common/src/wallet.rs`) | `BDKWalletManager` (`lampo-bdk-wallet`) | Keys, addresses, UTXOs, on-chain tx construction |
+| Composition | `LampoChainManager` (`lampod/src/chain/blockchain.rs`) | holds `Arc<dyn Backend>` + `Arc<dyn WalletManager>` | Glues LN ↔ BTC; implements LDK `FeeEstimator`/`Broadcaster`/`UtxoLookup` |
+
+**Invariant (hard):**
+
+> The on-chain wallet crate must not depend on LDK chain-sync types. Any coupling that LDK's `Listen`/`synchronize_listeners` requires lives **inside the concrete `Backend` implementation** (e.g. `lampo-chain` for bitcoind), never in the wallet and never as a cross-crate contract.
+
+**Corollary (backend swap is the priority axis):**
+
+> Unified sync is a property of the `Backend`, coordinated by a backend-agnostic `ChainSyncCoordinator`. The bitcoind backend implements it with `synchronize_listeners` + a local `Listen` adapter. A future Esplora/Electrum backend implements the *same* contract with transaction-based sync (`Confirm`). The wallet and the LN consumers are unaware of which backend is in use.
+
+This is the central revision relative to the original draft (and to ldk-node): the wallet exposes only **lampo-native** block/tx application; the bitcoind-specific `Listen` adapter is an implementation detail of `lampo-chain`.
 
 ## Problem statement
 
@@ -25,24 +47,36 @@ Lampo currently runs **two independent, competing chain-sync pipelines** over th
 - **Logs:** 0 ERROR/WARN; benign lock skip messages every 2 min.
 - **Anomaly:** no `Applied block` INFO lines and `bdk-wallet.db` mtime frozen while RPC scan advances (needs investigation during implementation).
 
+### Root cause
+
+Two pipelines, two RPC clients, no coordination:
+
+- `lampo-chain` (`LampoChainSync`) holds its own `lightning_block_sync::rpc::RpcClient` and runs `synchronize_listeners` (CM + ChainMonitor only) then `SpvClient::poll_best_tip()`.
+- `lampo-bdk-wallet` (`BDKWalletManager`) holds a **separate** `bdk_bitcoind_rpc::bitcoincore_rpc::Client` and runs an `Emitter` full-block scan on a `tokio-cron-scheduler` (every 2 min + 5s one-shot), independent of the LDK path.
+
+They contend for the same bitcoind, redundantly download the 190k→tip overlap, and starve each other. Neither knows the other's progress.
+
 ### User impact
 
-- Node appears “up” (`blockheight` = tip) but **Lightning chain listeners may not be caught up** and **on-chain wallet is not ready** for balances/UTXOs.
+- Node appears "up" (`blockheight` = tip) but **Lightning chain listeners may not be caught up** and **on-chain wallet is not ready** for balances/UTXOs.
 - Duplicate block downloads waste RPC bandwidth and wall-clock time.
-- Operators cannot distinguish “syncing” vs “ready” from `getinfo` alone.
+- Operators cannot distinguish "syncing" vs "ready" from `getinfo` alone.
 
 ## Goals
 
-1. **Single coordinated chain sync** for all LDK `Listen` targets including the on-chain wallet (ldk-node parity for bitcoind backend).
-2. **Serialize or prioritize** initial sync so LDK listener sync completes and is observable before/alongside wallet catch-up.
-3. **Expose sync progress** in `getinfo` (and logs): listener sync done, wallet scan height, percent/ETA optional.
-4. **Optional fast-sync / checkpoint jump** for signet and empty wallets (extends [#540](https://github.com/vincenzopalazzo/lampo.rs/issues/540)).
-5. **Retry/backoff** on transient RPC failures during `synchronize_listeners` (ldk-node pattern).
+1. **Single coordinated chain sync** that drives all chain consumers (LN channel manager, chain monitor, on-chain wallet) to tip over **one** RPC client.
+2. **Serialize / make observable** initial sync so listener sync completes (and is logged) before/alongside wallet catch-up.
+3. **Expose sync progress** in `getinfo` and logs.
+4. **Preserve the three trait seams** — wallet stays LDK-free; `Backend` stays swappable; coupling confined to the concrete backend crate. *(New, load-bearing goal.)*
+5. **Be ready for tx-based backends** (Esplora/Electrum) without touching the wallet. *(New.)*
+6. **Optional fast-sync / checkpoint jump** for signet and empty wallets (extends [#540](https://github.com/vincenzopalazzo/lampo.rs/issues/540)).
+7. **Retry/backoff** on transient RPC failures during sync (ldk-node pattern).
 
 ## Non-goals (this design)
 
 - Replacing Lampo with ldk-node or ldk-server.
-- Esplora/Electrum chain sources in the first PR (can follow; ldk-node tx-based sync is a separate follow-up).
+- **Implementing LDK `Listen`/`Confirm` directly on the wallet** (explicitly rejected — see Alternatives).
+- Building the Esplora/Electrum backend now (only making the seam ready).
 - Changing LDK persistence format for channel manager.
 - Mainnet policy for fast-sync defaults (must remain opt-in / safe).
 
@@ -53,42 +87,42 @@ sequenceDiagram
     participant CLI as lampod-cli
     participant W as lampo-bdk-wallet
     participant C as lampo-chain
-    participant RPC as Bitcoind RPC
+    participant RPC1 as bitcoind RPC (BDK client)
+    participant RPC2 as bitcoind RPC (LDK client)
 
     CLI->>W: wallet.listen() (cron + 5s oneshot)
-    CLI->>CLI: lampod.init()
+    CLI->>CLI: lampod.init(client)
     CLI->>CLI: lampod.listen()
     CLI->>C: spawn backend.listen()
 
-    par BDK path
-        W->>RPC: Emitter getblock loop (from wallet checkpoint)
-    and LDK path
-        C->>RPC: synchronize_listeners (CM + ChainMonitor only)
+    par BDK path (own client)
+        W->>RPC1: Emitter getblock loop (from wallet checkpoint ~21k)
+    and LDK path (own client)
+        C->>RPC2: synchronize_listeners (CM + ChainMonitor, from ~190k)
         Note over C: SpvClient poll after sync completes
     end
 ```
 
 ### LDK path (`lampo-chain/src/lib.rs`)
 
-- `listen()` spawns async task.
-- `synchronize_listeners` for **channel manager** and **chain monitor** only (not BDK wallet).
-- On success: log + `SpvClient::poll_best_tip()` loop.
-- Starting block: `channel_manager.current_best_block()` (e.g. 190537).
+- `LampoChainSync` owns an `Arc<RpcClient>` and implements both `BlockSource` and the lampo `Backend` trait.
+- `set_channel_manager` / `set_chain_monitor` inject the LDK objects via `OnceLock` (the wallet is **not** injected today).
+- `Backend::listen()` runs `synchronize_listeners` for **channel manager + chain monitor** from `channel_manager.current_best_block()` (e.g. 190537), logs, then runs an `SpvClient::poll_best_tip()` loop.
 
 ### On-chain path (`lampo-bdk-wallet/src/lib.rs`)
 
-- `sync()` uses `bdk_bitcoind_rpc::Emitter` from `wallet.latest_checkpoint()`.
+- `sync()` uses `bdk_bitcoind_rpc::Emitter` from `wallet.latest_checkpoint()` over a **separate** `bitcoincore_rpc::Client`.
 - Per-block `apply_block_connected_to` + `persist`.
-- Scheduled via `tokio-cron-scheduler` (every 2 min + 5s initial); holds `guard` for entire `sync()`.
+- Scheduled via `tokio-cron-scheduler` (every 2 min + 5s one-shot); holds a `guard` mutex for the entire `sync()`.
 - `reindex_from` only when `start_height == 0` or explicit `conf.reindex` forward jump ([#444](https://github.com/vincenzopalazzo/lampo.rs/issues/444)).
 
 ### Startup order (`lampod-cli/src/main.rs`)
 
-1. `wallet.listen().await?` — runs job scheduler (blocks until scheduler stops).
-2. `lampod.init()` — includes `channel_manager.listen()` (start CM, not chain backend).
-3. `run_httpd`, then `lampod.listen()` — spawns `onchain_manager().listen()` → chain backend.
+1. `wallet.listen().await?` — starts the job scheduler.
+2. `lampod.init(client)` — wires onchaind/channeld/etc., injects handler + CM + chain monitor into the backend.
+3. `run_httpd`, then `lampod.listen()` — spawns `onchain_manager().listen()` → `Backend::listen()`.
 
-**Issue:** wallet scheduler can start BDK full scan **before or in parallel with** chain `synchronize_listeners`, both hammering the same RPC client.
+**Issue:** wallet scheduler can start the BDK full scan **before or in parallel with** the LDK `synchronize_listeners`, both hammering bitcoind over distinct clients.
 
 ## Reference architecture (ldk-node / ldk-server)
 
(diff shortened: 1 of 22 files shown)


lampod/src/jsonrpc/inventory.rs
```diff
@@ -54,11 +55,33 @@ pub async fn json_getinfo(ctx: &LampoDaemon, request: &json::Value) -> Result<js
...
 
+/// Block until the node's initial chain sync (LDK listeners + on-chain wallet)
+/// completes, then return the final sync status. Useful for scripts and CI.
+pub async fn json_sync_wallets(
+    ctx: &LampoDaemon,
+    request: &json::Value,
+) -> Result<json::Value, Error> {
+    log::info!("calling `sync_wallets` with request `{:?}`", request);
+    let chain_sync = ctx.chain_sync();
+    chain_sync.wait_initial_sync_complete().await;
```
vincenzopalazzo: **Medium:** `wait_initial_sync_complete()` only resolves on `Running`. If the node never reaches it (e.g. the Emitter errors before its first drain, or a misconfigured backend), this handler blocks forever and pins an actix worker. Can we bound it with a timeout that returns the current `SyncStatus` (with `initial_sync_complete=false`) instead of hanging the connection?

lampo-common/src/model/getinfo.rs
```diff
@@ -13,6 +13,14 @@ pub struct GetInfo {
     pub address: Vec<NetworkInfo>,
     pub block_hash: String,
     pub wallet_height: u64,
+    /// Live wallet scan height during initial catch-up, if known.
+    pub wallet_scan_height: Option<u32>,
```
vincenzopalazzo, line 18: **Medium:** `getinfo` is a public response, but the four new fields have no `#[serde(default)]`. A newer client deserializing an *older* daemon's reply (missing these keys) would fail to parse. Add the attribute to each new field:

```suggestion
    #[serde(default)]
    pub wallet_scan_height: Option<u32>,
```

...and the same for `chain_listeners_synced`, `initial_sync_complete`, and `sync_in_progress`.

lampo-common/src/chainsync.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    /// Report the wallet's current scan height (live progress during catch-up).
+    pub fn set_wallet_scan_height(&self, height: u32) {
+        self.wallet_scan_height
+            .store(u64::from(height), Ordering::Relaxed);
+    }
+
+    /// Resolve once the chain listeners have synced (state is `ListenersSynced`
+    /// or `Running`). Returns immediately if already past `PendingInitialSync`.
+    pub async fn wait_listeners_synced(&self) {
```
vincenzopalazzo: **Cleanup:** `wait_listeners_synced` is unused — `pub`, so no dead-code warning, but nothing calls it. I added it in the first commit expecting a startup gate, but the gate ended up wallet-side (the cron checks `chain_listeners_synced()`). Either wire it into the startup path or drop it; `wait_initial_sync_complete` is the one actually used.

tools/check-wallet-decoupling.sh
```diff
@@ -0,0 +1,25 @@
...
+# The LDK <-> wallet coupling lives in `lampo-chain` via the
+# `WalletChainListener` adapter, so the wallet stays independently replaceable.
+# See docs/designs/unified-chain-sync.md.
+#
+# Note: a *transitive* `lightning` dep is expected (through `lampo-common`); we
+# only forbid a *direct* dependency, hence `--depth 1`.
+set -euo pipefail
+
+direct=$(cargo tree -p lampo-bdk-wallet --edges normal --depth 1 --prefix none 2>/dev/null \
+    | grep -iE 'lightning' || true)
```
vincenzopalazzo: **Cleanup:** `grep -iE 'lightning'` is a case-insensitive substring match, so a direct dep with `lightning` anywhere in its name would false-positive, and `-i` is unnecessary for crate names. Anchoring to the LDK crates we actually forbid is tighter:

```suggestion
    | grep -E '^lightning' || true)
```

lampo-common/src/chainsync.rs
```diff
@@ -0,0 +1,185 @@
...
+                *state = SyncState::ListenersSynced;
+                true
+            } else {
+                false
+            }
+        });
+    }
+
+    /// Advance to `Running`: the initial sync is fully complete.
+    pub fn mark_running(&self) {
```
vincenzopalazzo, line 78: **Suggestion:** `mark_running` uses `send_replace(Running)` unconditionally, so a stray call from `PendingInitialSync` would skip `ListenersSynced`. The only caller guards with `chain_listeners_synced()` today, so it's safe — but a defensive guard here (advance to `Running` only from `ListenersSynced`/`Running`) makes the state machine self-consistent regardless of caller. Non-blocking.

lampo-chain/src/lib.rs
```diff
@@ -10,13 +10,59 @@ use lightning_block_sync::{BlockSource, SpvClient};
...
+    fn block_connected(&self, block: &Block, height: u32) {
+        if let Err(err) = self.wallet.apply_block(block, height) {
+            log::error!(
+                target: "lampo-chain",
+                "on-chain wallet apply_block at height {height} failed: {err}"
+            );
+        }
+    }
+
+    fn blocks_disconnected(&self, _fork_point_block: chain::BlockLocator) {
```
vincenzopalazzo, line 83: **Suggestion:** `blocks_disconnected` is a no-op, relying on BDK reconnecting via the next `block_connected`'s `connected_to`. I think this is correct for the catch-up path (no deep reorgs far below tip) and ongoing reorgs are handled by the gated Emitter — but it's untested for a deep reorg *during* `synchronize_listeners`. The deferred signet integration test (design doc PR plan) should cover this before PR 5 routes ongoing updates through this adapter.

vincenzopalazzo (commented):
Self-review: I found 6 issues (2 medium, none blocking) across the coordinator, the `getinfo` schema, and the new `sync_wallets` RPC. The inline comments capture the specifics. CI is green including the integration suite (`test`, regtest+cln).

## Code Review

### Positive Aspects
- The `WalletChainListener` adapter (`lampo-chain/src/lib.rs:30`) confines all LDK<->BDK coupling to the bitcoind backend crate, so `lampo-bdk-wallet` keeps a zero `lightning*` dependency — now enforced by the CI guard. This is the whole modularity goal, made mechanical.
- The `std::sync::Mutex` switch (`wallet: synchronous chain-state locking`) makes the wallet apply runtime-agnostic *and* lets the compiler enforce "no guard across `.await`" — which is what surfaced and fixed the latent `listen()` guard-shadow bug (the frozen-`bdk-wallet.db` symptom from #545).
- Reusing `synchronize_listeners` for the wallet's initial catch-up dedups the overlapping block range with the channel manager / chain monitor in a single RPC pass, instead of a second `getblock` scan.

---

### Issues & Suggestions

#### 1. `sync_wallets` can block indefinitely (Medium)
`json_sync_wallets` awaits `wait_initial_sync_complete()`, which only resolves on `Running`. If that state is never reached, the HTTP handler hangs forever and pins an actix worker. See inline on `lampod/src/jsonrpc/inventory.rs:75` — can we bound it with a timeout that returns the current `SyncStatus`?

#### 2. New `getinfo` fields lack `#[serde(default)]` (Medium)
`getinfo` is a public response; a newer client reading an older daemon would fail to deserialize. Inline on `lampo-common/src/model/getinfo.rs:17`.

#### 3. `wait_listeners_synced` is unused (Cleanup)
Dead public API — added anticipating a startup gate that ended up wallet-side. Inline on `lampo-common/src/chainsync.rs:92`.

#### 4. Decoupling-guard grep is loose (Cleanup)
`grep -iE 'lightning'` is a case-insensitive substring match. Inline on `tools/check-wallet-decoupling.sh:15`.

#### 5. `mark_running` has no state guard (Suggestion)
Relies on the caller checking `chain_listeners_synced()`. Inline on `lampo-common/src/chainsync.rs:72`.

#### 6. `blocks_disconnected` no-op is untested for deep reorgs (Suggestion)
Inline on `lampo-chain/src/lib.rs:58`. Correct for the catch-up path; the deferred signet test should exercise it before PR 5.

Non-blocking nit (body only): the deprecated `LampoInventoryManager::get_info_node` hardcodes `sync_in_progress: true` rather than deriving from the coordinator like the live path. Harmless since the method is `#[deprecated]` and unused, but it'll drift — worth a comment or deletion in a follow-up.

---

### Summary
Self-review: 6 findings, **none blocking**. The two mediums (RPC indefinite block, `serde(default)`) are worth folding into this PR; the rest are follow-ups. The implementation is sound and CI-green end-to-end. PR 5 (retire the Emitter for ongoing) is deliberately deferred — its core objective is already met by PR 2 + PR 4 (see the design doc's PR plan).

---

### Decision Log

**Hardest decision:** the sync->async bridge for the `Listen` adapter (Q4). `chain::Listen::block_connected` is synchronous, but the wallet apply was `async` (tokio mutex). I chose to make the wallet's chain-application a *synchronous core* (`std::sync::Mutex` + sync `apply_block`) rather than bridge with `block_in_place`/`Handle::block_on`. The deciding factor: `LampoDaemon` is pitched as an embeddable library (`lampod/src/lib.rs:80`), and `block_in_place` requires a multi-thread runtime and panics on a current-thread one — a footgun for embedders. The sync core is runtime-agnostic and the compiler enforces correctness.

**Alternatives rejected:**
- `block_in_place` + `Handle::block_on` inside `block_connected`: smaller diff, but runtime-flavor fragile and unverifiable in the integration tests I can't run locally.
- `impl chain::Listen` directly on `BDKWalletManager` (literal ldk-node parity): punctures the `WalletManager` seam and would block the Esplora/Electrum backend swap — the exact thing this design protects.
- Full PR 5 (retire the Emitter, drive ongoing from the `SpvClient` poll): needs a custom 3-listener `Listen` impl (LDK only impls it for 2-tuples of `Deref`) with real reorg/ongoing risk and no functional gain over the gated incremental Emitter. Deferred with rationale.

**Least confident about:** wallet reorg handling during `synchronize_listeners` — `blocks_disconnected` is a no-op that leans on BDK reconnecting via the next `block_connected`'s `connected_to`. I'm confident for the common catch-up case (no deep reorgs far below tip) and ongoing reorgs are still handled by the Emitter, but I'd want the deferred signet integration test to prove the deep-reorg path before PR 5 routes ongoing updates through this adapter. Secondary worry: mixed-version `getinfo` deserialization (finding #2).

lampod/src/jsonrpc/inventory.rs
vincenzopalazzo: **Fixed in 95bb89b.** Bounded with a 600s timeout; returns the current `SyncStatus` on timeout so the caller re-polls instead of pinning the worker.

lampo-common/src/model/getinfo.rs
vincenzopalazzo: **Fixed in 95bb89b.** Added `#[serde(default)]` to all four new fields.

lampo-common/src/chainsync.rs
vincenzopalazzo: **Fixed in 95bb89b.** Removed the unused method; `wait_initial_sync_complete` is the one actually in use.

tools/check-wallet-decoupling.sh
vincenzopalazzo: **Fixed in 95bb89b.** Anchored the grep to `^lightning`.

lampo-common/src/chainsync.rs
vincenzopalazzo: **Fixed in 95bb89b.** `mark_running` now only advances from `ListenersSynced` (no-op otherwise), with a unit test.

lampo-chain/src/lib.rs
vincenzopalazzo: Deferred to the signet integration test (design doc PR plan). The no-op is correct for the catch-up path; deep reorgs get validated there before PR 5.

lampod/src/jsonrpc/inventory.rs
```diff
@@ -54,6 +55,10 @@ pub async fn json_getinfo(ctx: &LampoDaemon, request: &json::Value) -> Result<js
         lampo_dir,
         address: address_vec,
         wallet_height: wallet_tips.to_consensus_u32() as u64,
+        wallet_scan_height: chain_sync.wallet_scan_height(),
+        chain_listeners_synced: chain_sync.chain_listeners_synced(),
+        initial_sync_complete: chain_sync.initial_sync_complete(),
+        sync_in_progress: chain_sync.sync_in_progress(),
```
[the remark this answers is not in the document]
vincenzopalazzo: **Resolved by later commits.** The coordinator is now driven: the backend marks `ListenersSynced` after `synchronize_listeners` and the wallet reports scan height + `Running` on catch-up. By the PR head, `getinfo` reflects real state.

lampod/src/lib.rs
```diff
@@ -115,9 +117,16 @@ impl LampoDaemon {
             offchain_manager: None,
             handler: None,
             shutdown: Arc::new(AtomicBool::new(false)),
+            chain_sync: Arc::new(ChainSyncCoordinator::new()),
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in 95bb89b.** Good call -- moved `wallet.set_coordinator(...)` into `LampoDaemon::new`, so every daemon (CLI, tests, embedders) wires it consistently; dropped the per-caller calls.

lampo-bdk-wallet/src/lib.rs
```diff
@@ -369,11 +393,51 @@ impl WalletManager for BDKWalletManager {
...
+            // Gate: hold the Emitter scan until the LDK chain listeners have
+            // synced, so the two pipelines don't compete for the same RPC.
+            // Active only when a coordinator is wired in and
+            // `wallet_sync_parallel` is unset; otherwise sync runs as before.
+            let parallel = wallet.conf.wallet_sync_parallel.unwrap_or(false);
+            if !parallel {
+                if let Some(coordinator) = wallet.coordinator.get() {
+                    if !coordinator.chain_listeners_synced() {
+                        log::info!(target: "lampo-wallet", "Waiting for chain listeners to sync before scanning the wallet");
+                        return Ok(());
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in 95bb89b.** Good catch -- `synchronize_listeners` now retries with backoff, so a transient RPC failure no longer leaves the coordinator stuck in `PendingInitialSync` (which would permanently gate the wallet).

lampo-chain/src/lib.rs
```diff
@@ -10,13 +10,59 @@ use lightning_block_sync::{BlockSource, SpvClient};
...
+        // Lampo's bitcoind backend delivers full blocks, so `block_connected`
+        // below is what actually runs. Mirrors ldk-node's wallet listener.
+        debug_assert!(
+            false,
+            "filtered_block_connected is unsupported for the on-chain wallet"
+        );
+    }
+
+    fn block_connected(&self, block: &Block, height: u32) {
+        if let Err(err) = self.wallet.apply_block(block, height) {
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in 95bb89b.** The Listen adapter now records apply failures and `listen()` warns after the pass; the gated Emitter recovers the wallet from its last good checkpoint instead of the failure being swallowed.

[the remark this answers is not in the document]
vincenzopalazzo: Same as above -- **fixed in 95bb89b.**

lampo-chain/src/lib.rs
```diff
@@ -231,7 +297,25 @@ impl Backend for LampoChainSync {
...
+        // the overlapping block range instead of running a second `getblock`
+        // scan. Kept in locals so the listener outlives the `chain_listeners`
+        // vec consumed by `synchronize_listeners`.
+        let unified = self
+            .config
+            .sync_mode
+            .as_deref()
+            .map(|mode| !mode.eq_ignore_ascii_case("legacy"))
+            .unwrap_or(true);
+        let wallet = if unified { self.wallet() } else { None };
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in 95bb89b.** `unified` now requires `!wallet_sync_parallel`, so parallel mode excludes the wallet from `synchronize_listeners` -- the two paths are mutually exclusive.

lampo-common/src/chainsync.rs
```diff
@@ -0,0 +1,195 @@
+//! Backend-agnostic chain-sync coordination.
+//!
+//! `ChainSyncCoordinator` is a small, dependency-free state holder that lets
+//! the chain backend, the on-chain wallet, and the JSON-RPC layer agree on
+//! where the node is in its initial sync — without any of them depending on
+//! each other. It carries **no** LDK or BDK types on purpose: the concrete
+//! `Backend` drives the transitions, the wallet reports progress, and
+//! `getinfo` reads the state. See `docs/designs/unified-chain-sync.md`.
+//!
+//! In this first PR the coordinator is wired in but not yet driven; later PRs
```
vincenzopalazzo: **Cleanup:** Stale module doc — "In this first PR the coordinator is wired in but not yet driven; later PRs call mark_listeners_synced / mark_running". This PR *does* drive it (`mark_listeners_synced` from `lampo-chain`, `mark_running` from the wallet's `sync()`). Suggest:

```suggestion
//! The concrete `Backend` calls `mark_listeners_synced` after its initial
//! sync pass; the wallet calls `mark_running` once its scan completes.
```

vincenzopalazzo (commented):
Self-review: Concept ACK — good to merge, nothing blocking. Two doc inaccuracies to fix (one stale module comment, one overclaiming table row) and a few non-blocking suggestions below; inline comments have the specifics. Answering the two questions directly: **(1) yes, good to merge** — CI green and I verified the fix live on the signet box; **(2) no, this does not break Lampo's design** — it enforces it, and it now has a CI guard to prove it.

Production Readiness: GO

## Code Review

### Does this break Lampo's design?

No — the opposite. The load-bearing decision is to *not* copy ldk-node's `impl Listen for Wallet`, and instead confine the unavoidable LDK↔wallet coupling to the bitcoind backend crate:

- The three seams stay intact. `Backend` (`lampo-common/src/backend.rs`) gains only default no-op methods (`set_coordinator`, `set_wallet_manager`), so any other backend keeps compiling unchanged. `WalletManager` (`lampo-common/src/wallet.rs`) gains `current_best_block`/`apply_block` in pure `bitcoin` types — no LDK. The composition in `lampod` stays backend-agnostic.
- `WalletChainListener` (`lampo-chain/src/lib.rs:31`) is the only place LDK chain-sync types meet the wallet — exactly the "backend-confined coupling" principle from the design doc.
- The invariant is now enforced, not just documented: `tools/check-wallet-decoupling.sh` runs in `build.yml` and fails CI if `lampo-bdk-wallet` ever takes a direct `lightning*` dependency.
- The one real design commitment is "the backend drives the wallet" (alternative 5 — hoisting orchestration into `lampod` — explicitly rejected in the doc). That is consistent with Lampo's backend-selected architecture and keeps `lampod` free of bitcoind-specific glue. Make sense to me.

One honest caveat: `lampo-common` itself still depends on LDK, so "the wallet is LDK-free" really means "no *direct* LDK dep, and no LDK chain-sync types on the wallet's API surface". That is the level of purity that matters for replaceability, and the guard checks exactly that — just worth being precise about what the invariant is.

### Positive aspects

- The std-mutex wallet core (`lampo-bdk-wallet/src/lib.rs`) is the right call and kills two birds: it fixes the guard-shadow bug that froze `bdk-wallet.db` in production (the missing `Applied block` logs), *and* makes the wallet drivable from sync `Listen` on any runtime flavor with no `block_on` bridge.
- The coordinator (`lampo-common/src/chainsync.rs`) is genuinely dependency-free and ships with real state-machine unit tests (+5 tests in a repo that had ~15 total).
- The retry loop rebuilds the listener set from current best blocks on each attempt, so a mid-pass RPC failure resumes where it left off instead of restarting the catch-up.
- Runtime verification on the Ocean signet box (PR head `95bb89b`, Bitcoin Core v31.1 backend): clean sync to tip 315016 with **zero** `Blocks must be connected in chain-order` panics, and `getinfo` reports `chain_listeners_synced=true`, `initial_sync_complete=true`, `sync_in_progress=false`, `wallet_scan_height=315016`. `main` panics in this exact setup, so the fix is confirmed against a live backend, not just in theory.

### Issues & Suggestions

The inline comments carry the line-level items. Summary of what's there:

#### 1. Stale module doc in chainsync.rs (Cleanup)
`lampo-common/src/chainsync.rs:10` still says "in this first PR the coordinator is wired in but not yet driven" — this PR drives it.

#### 2. Doc table overclaims "RPC clients: 1" (Medium)
`docs/designs/unified-chain-sync.md:155` — implementation still runs 2 clients; the gated Emitter serves ongoing/mempool updates until PR 5 lands.

#### 3. `blocks_disconnected` no-op relies on the Emitter as safety net (Medium, non-blocking)
`lampo-chain/src/lib.rs:83` — fine today, but the regtest reorg test must land before PR 5 retires the Emitter.

#### 4. No unit test for the adapter (Medium, non-blocking)
The doc's testing plan promised adapter delegation tests; not delivered. CI green + my signet run cover the happy path, but this code exists to fix a chain-order panic — it deserves a mock-based test.

#### 5. Fixed 5s retry backoff (Suggestion)
`lampo-chain/src/lib.rs:401` — ldk-node uses exponential backoff; fine for a first cut.

#### 6. Unknown `sync-mode` value silently becomes unified (Suggestion)
`lampo-chain/src/lib.rs:338` — a conf typo silently changes sync strategy; validate at startup.

#### 7. `sync_wallets` doc vs 600s bound (Cleanup)
`lampod/src/jsonrpc/inventory.rs:67` — doc says "block until completes", behavior is bounded and returns current status on timeout.

#### 8. `set_coordinator` panics on second set (Suggestion)
`lampo-bdk-wallet/src/lib.rs:320` — consistent with the existing OnceLock style, but an embedder building two daemons over one wallet panics.

#### 9. Decoupling guard only checks normal edges (Cleanup, follow-up)
`tools/check-wallet-decoupling.sh:14` — a dev-dependency on `lightning*` would slip through.

---

### Summary

Concept ACK — good to merge. Fix the two doc nits here or in a quick follow-up; everything else is follow-up material that belongs with PR 5 (retire the Emitter) and 8b (signet integration test). Design-wise this is the right shape: it gets ldk-node's single-pass efficiency *without* puncturing the `WalletManager`/`Backend` seams, and the CI guard means future coupling creep fails the build instead of being discovered in a review.

---

### Decision Log

**Hardest decision:** rejecting literal ldk-node parity (`impl Listen` on the wallet) and instead making the wallet core *synchronous* (std mutexes, PR 3b) so a sync `Listen` adapter inside `lampo-chain` can drive it on any runtime flavor. The sync-core choice is what makes the whole decoupled design work without a `block_on` bridge — and as a bonus it fixed the guard-shadow freeze we saw in production. The risk accepted: std mutexes in async code paths, mitigated by keeping critical sections to short BDK apply + persist.

**Alternatives rejected:**
- `impl Listen`/`Confirm` directly on the BDK wallet (ldk-node parity): punctures the `WalletManager` seam and hard-codes the block-based path, obstructing the Esplora/Electrum swap, which is the priority axis.
- Hoisting `synchronize_listeners` orchestration into `lampod`: viable, but spreads bitcoind-specific LDK glue into the composition layer.
- Retiring the Emitter in this PR: LDK only implements `Listen` for `(T, U)` pairs, so a 3-listener tuple needs a custom `Listen` impl — real reorg/ongoing-update risk, better landed with the signet integration test.

**Least confident about:** reorg behavior at the tip during the unified pass, given `blocks_disconnected` is a no-op and recovery relies on BDK rolling back via the next `block_connected`'s `connected_to`. During historical catch-up this is safe; at the tip boundary the gated Emitter is the recovery path. The regtest reorg test (doc testing-plan item 4) is the thing I most want before PR 5 removes that safety net.

docs/designs/unified-chain-sync.md
```diff
@@ -53,244 +87,381 @@ sequenceDiagram
...
-| RPC passes on restart | 2 (overlapping ranges) | 1 initial + incremental poll |
-| Initial sync completion | Wallet: no API flag; LDK: log line missing in prod | Log + metrics timestamps |
-| RPC error handling | Emitter error logged; listener stall unclear | Retry/backoff on `synchronize_listeners` |
-| Progress in API | `wallet_height` stale during scan | `current_best_block`, sync timestamps |
-| Fast sync | `reindex` / height 0 only | New wallet → tip checkpoint; restored → per-listener best block in one sync |
-| Startup coupling | Wallet cron independent of LDK sync | `start()` coordinates background sync |
+| Aspect | Lampo (today) | ldk-node (bitcoind) | Lampo (this design) |
+|--------|---------------|---------------------|---------------------|
+| Listener set | CM + ChainMonitor | Wallet + CM + Sweeper + ChainMonitor | CM + ChainMonitor + **wallet adapter** (bitcoind backend only) |
+| RPC clients | 2 (separate) | 1 | 1 |
```
vincenzopalazzo: **Medium:** This row overclaims. The implementation still runs **2 RPC clients**: the wallet keeps its own `bitcoincore_rpc::Client` (`lampo-bdk-wallet/src/lib.rs:166`), and the gated `Emitter` remains the path for ongoing/mempool updates — PR 5 (retire the Emitter) is explicitly deferred below. The "one RPC stream" property holds for the *initial* catch-up only. Either annotate the row (`1 during initial sync; 2 until PR 5`) or footnote it, otherwise the doc promises something the code doesn't deliver yet.

lampo-chain/src/lib.rs
```diff
@@ -10,13 +11,83 @@ use lightning_block_sync::{BlockSource, SpvClient};
...
+        if let Err(err) = self.wallet.apply_block(block, height) {
+            log::error!(
+                target: "lampo-chain",
+                "on-chain wallet apply_block at height {height} failed: {err}"
+            );
+            self.failed.store(true, Ordering::Relaxed);
+        }
+    }
+
+    fn blocks_disconnected(&self, _fork_point_block: chain::BlockLocator) {
```
vincenzopalazzo, line 83: **Medium (non-blocking):** The no-op here is fine *today* — BDK rolls the wallet chain back via the next `block_connected`'s `connected_to`, and reorgs don't occur during historical catch-up. But a tip-boundary reorg mid-pass leaves the wallet diverged until the gated Emitter recovers it. That makes the Emitter the safety net for this case, so the regtest reorg test (doc testing-plan item 4) really needs to land **before** PR 5 retires the Emitter. Can we make that ordering explicit in the PR plan?

lampo-chain/src/lib.rs
```diff
@@ -10,13 +11,83 @@ use lightning_block_sync::{BlockSource, SpvClient};
...
+use lampo_common::wallet::WalletManager;
+
+/// Adapts the on-chain wallet to LDK's [`chain::Listen`] so it can ride the
+/// same `synchronize_listeners` pass as the channel manager and chain monitor
+/// -- one RPC stream for the whole node, instead of a second `getblock` scan.
+///
+/// This is the *only* place the LDK chain-sync types meet the wallet: the
+/// coupling is confined to the bitcoind backend crate. The wallet itself
+/// stays free of LDK, driven through the lampo-native `WalletManager` API.
+struct WalletChainListener {
```
vincenzopalazzo, line 31: **Medium (non-blocking):** No unit test for the adapter. The design doc's testing plan item 1 promised "adapter `block_connected`/`block_disconnected` delegate to `WalletManager`" — not delivered. Given this code exists specifically to fix a chain-order panic, a delegation test with a mock `WalletManager` (assert apply order, assert `failed` flag set on error) is cheap insurance. CI green + my live signet run cover the happy path, but CI should own this regression, not a manual run.

lampo-chain/src/lib.rs
```diff
@@ -249,13 +348,80 @@ impl Backend for LampoChainSync {
...
                 .await
-                .map_err(|e| error::anyhow!("Failed to synchronize chain listeners: {:?}", e))?;
+            {
+                Ok(result) => break result,
+                Err(e) => {
+                    log::error!(
+                        target: "lampo-chain",
+                        "Failed to synchronize chain listeners, retrying in 5s: {:?}", e
+                    );
+                    tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
vincenzopalazzo: **Suggestion:** Fixed 5s retry. ldk-node uses exponential backoff for this loop; consider backoff with a cap (e.g. 5s → 60s) plus a periodic WARN so a permanently-stuck node is visible in logs instead of emitting the same ERROR line every 5 seconds. Non-blocking for this PR.

lampo-chain/src/lib.rs
```diff
@@ -231,16 +322,24 @@ impl Backend for LampoChainSync {
...
+        // `getblock` scan. Excluded in `legacy` sync mode, and when
+        // `wallet_sync_parallel` is set (the wallet then runs its own Emitter,
+        // so attaching it here too would double-apply blocks). Kept in a local
+        // so it outlives the `chain_listeners` vec consumed on each attempt.
+        let parallel = self.config.wallet_sync_parallel.unwrap_or(false);
+        let legacy = self
+            .config
+            .sync_mode
+            .as_deref()
+            .map(|mode| mode.eq_ignore_ascii_case("legacy"))
```
vincenzopalazzo: **Suggestion:** An unrecognized `sync-mode` value silently becomes `unified` — a typo in the conf (`sync-mode=unfiied`) silently changes the node's sync strategy. Can we validate at startup and refuse (or at least WARN) on anything outside `{unified, legacy}`? Same parsing question applies to the conf layer.

lampod/src/jsonrpc/inventory.rs
```diff
@@ -54,11 +55,44 @@ pub async fn json_getinfo(ctx: &LampoDaemon, request: &json::Value) -> Result<js
...
     };
 
     Ok(json::to_value(getinfo)?)
 }
 
+/// Block until the node's initial chain sync (LDK listeners + on-chain wallet)
```
vincenzopalazzo: **Cleanup:** Doc says "Block until the node's initial chain sync completes" but the wait is bounded at 600s and returns the current (possibly incomplete) status on timeout — the code comment below explains it, but the doc comment should say it too, since callers will rely on this:

```suggestion
/// Wait (up to 600s) for the node's initial chain sync (LDK listeners +
/// on-chain wallet) to complete, then return the sync status. On timeout
/// returns the current, possibly incomplete, status; callers should re-poll.
```

lampo-bdk-wallet/src/lib.rs
```diff
@@ -272,6 +314,12 @@ impl WalletManager for BDKWalletManager {
         Ok(txs)
     }
 
+    fn set_coordinator(&self, coordinator: Arc<ChainSyncCoordinator>) {
+        self.coordinator
+            .set(coordinator)
+            .unwrap_or_else(|_| panic!("chain sync coordinator already set"));
```
vincenzopalazzo: **Suggestion:** Panicking on second set is consistent with the existing `OnceLock` style in this codebase, but note `LampoDaemon::new` now wires the wallet unconditionally — an embedder that constructs two daemons over the same wallet (or wires the wallet itself first) panics at runtime. Consider making it idempotent (`let _ = ...set(...)` + debug log) or documenting the single-wiring requirement on `WalletManager::set_coordinator`.

tools/check-wallet-decoupling.sh
```diff
@@ -0,0 +1,25 @@
...
+#
+# The LDK <-> wallet coupling lives in `lampo-chain` via the
+# `WalletChainListener` adapter, so the wallet stays independently replaceable.
+# See docs/designs/unified-chain-sync.md.
+#
+# Note: a *transitive* `lightning` dep is expected (through `lampo-common`); we
+# only forbid a *direct* dependency, hence `--depth 1`.
+set -euo pipefail
+
+direct=$(cargo tree -p lampo-bdk-wallet --edges normal --depth 1 --prefix none 2>/dev/null \
```
vincenzopalazzo, line 17: **Cleanup (follow-up):** `--edges normal` only — a *dev-dependency* on `lightning*` in `lampo-bdk-wallet` would slip through the guard. Probably nobody does that deliberately, but since this script is the enforcement point for the design invariant, `--edges normal,dev` (or a comment saying why dev is excluded) would make it airtight.

lampo-common/src/chainsync.rs
vincenzopalazzo: Fixed in 438e1d4: updated the module doc to state the coordinator is now driven from chain and wallet.

docs/designs/unified-chain-sync.md
vincenzopalazzo: Fixed in 438e1d4: annotated the RPC clients row as 1 during initial sync; 2 until PR 5.

lampo-chain/src/lib.rs
vincenzopalazzo: Fixed in 438e1d4: added an explicit ordering note in the PR plan that the regtest reorg test (8b) must land before PR 5.

lampo-chain/src/lib.rs
vincenzopalazzo: Fixed in 438e1d4: added a unit test module with a mock WalletManager covering ordered delegation and failure flag.

lampo-chain/src/lib.rs
vincenzopalazzo: Fixed in 438e1d4: replaced the fixed 5s retry with exponential backoff capped at 60s.

lampo-chain/src/lib.rs
vincenzopalazzo: Fixed in 438e1d4: added a WARN log for unrecognized sync_mode values, falling back to unified.

lampod/src/jsonrpc/inventory.rs
vincenzopalazzo: Fixed in 438e1d4: corrected the doc comment to describe the bounded 600s wait and timeout behavior.

lampo-bdk-wallet/src/lib.rs
vincenzopalazzo: Fixed in 438e1d4: made set_coordinator idempotent with a debug log; also applied to chain set_coordinator and set_wallet_manager.

tools/check-wallet-decoupling.sh
vincenzopalazzo: Fixed in 438e1d4: added a comment explaining why only --edges normal is checked.

lampo-common/src/chainsync.rs
vincenzopalazzo: Fixed in 438e1d4: updated stale module doc to describe the backend and wallet coordinator calls.

docs/designs/unified-chain-sync.md
vincenzopalazzo: Fixed in 438e1d4: annotated the RPC-client comparison as 1 during initial sync and 2 until PR 5.

lampo-chain/src/lib.rs
vincenzopalazzo: Fixed in 438e1d4: added explicit PR-plan ordering so the regtest reorg test lands before PR 5.

lampo-chain/src/lib.rs
vincenzopalazzo: Fixed in 438e1d4: added WalletChainListener unit tests with a mock wallet.

lampo-chain/src/lib.rs
vincenzopalazzo: Fixed in 438e1d4: switched synchronize_listeners retry to exponential backoff capped at 60s.

lampo-chain/src/lib.rs
vincenzopalazzo: Fixed in 438e1d4: added a WARN log for unrecognized sync_mode values while keeping unified fallback.

lampod/src/jsonrpc/inventory.rs
vincenzopalazzo: Fixed in 438e1d4: corrected sync_wallets doc to note the 600s bounded wait and re-poll behavior.

lampo-bdk-wallet/src/lib.rs
vincenzopalazzo: Fixed in 438e1d4: made set_coordinator/set_wallet_manager idempotent instead of panicking.

tools/check-wallet-decoupling.sh
vincenzopalazzo: Fixed in 438e1d4: added a comment explaining --edges normal in the decoupling check.

vincenzopalazzo:
Closing this in favor of #551, that replaces #546, #548 and #549 with a single series on top of the current main. Same content, cleaner history, and the review feedback commits are folded into the commits they belong to!

PR #548, opened 2026-07-06
Fix(sync): wire chain-sync coordinator into the on-chain manager

vincenzopalazzo:
## Summary

Fixes the unreachable `mark_listeners_synced` call in the unified chain-sync path (#545 / #546).

- **What was broken:** the on-chain wallet cron is gated behind `chain_listeners_synced`, which is only advanced by `mark_listeners_synced()`. That call lives in `LampoChainManager::listen()` (guarded by `if let Some(coordinator) = self.coordinator.get()`), but the coordinator was injected into the wallet (`LampoDaemon::new`) and the backend `client` (`LampoDaemon::init`) — **never into the `LampoChainManager`** itself. `set_coordinator` on the on-chain manager had zero callers, so `self.coordinator` was always `None`.
- **Effect:** the coordinator stayed `PendingInitialSync` forever; the wallet cron logged `Waiting for chain listeners to sync` every 2 minutes and made zero progress, even though the listeners were running. Verified live on signet: a restored wallet (height 21021 vs tip 312037) showed `chain_listeners_synced: false` / `wallet_scan_height: null` indefinitely after 4+ minutes.
- **Root cause:** missing coordinator wiring on the on-chain manager.
- **Fix:** also call `set_coordinator` on the on-chain manager in `LampoDaemon::init`, so `mark_listeners_synced()` can fire once `synchronize_listeners` completes.

`set_coordinator` is a `Backend` trait method, so no visibility change is needed — it was simply never invoked on the on-chain manager.

## Test plan

- [x] `cargo check --workspace --all-targets` clean (only pre-existing warnings)
- [x] `cargo test -p lampo-common chainsync` — all 5 coordinator unit tests pass
- [x] Deployed to the signet host and confirmed the node boots with the coordinator now wired to the on-chain manager

## Known follow-up (not addressed here — needs a design decision)

While verifying this fix live, a second, deeper issue surfaced on top of the wiring bug. After the wiring fix the coordinator is reachable, but on this host `synchronize_listeners` (which drives `mark_listeners_synced`) **hangs** rather than completing:

- the on-chain `BlockSource` uses `lightning_block_sync::rpc::RpcClient` (a separate, unlogged client from the wallet's `bitcoincore_rpc`), so its calls don't appear in the RPC histogram;
- the lampo process holds an `ESTABLISHED` but **idle** connection to the backend at `0.4% CPU` — `synchronize_listeners` is blocked awaiting an RPC response that never returns (the SSH-tunneled backend shows `CLOSE-WAIT`/`FIN-WAIT-2` sockets);
- because the wallet cron is **hard-gated** behind `chain_listeners_synced` with no timeout/escape path, a hung listener sync permanently deadlocks the wallet (zero `getblock` progress), whereas `main` scans the wallet in parallel and progresses.

This is precisely the P2 concern the review raised ("…needs an escape path when the backend sync has failed"). The wiring fix in this PR is necessary regardless (without it `mark_listeners_synced` is unreachable even on a healthy backend), but fully resolving the deadlock will additionally require either a bounded timeout / escape hatch in the wallet gate, or a read timeout on the `RpcClient`. I've left that as a separate change for the maintainer to decide on, with the live evidence captured here.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -88,3 +88,8 @@ jobs:
             exit 1
           fi
           echo "OK: each lightning* crate resolves to a single version."
+
+      # Guards the unified-chain-sync invariant (#545): the on-chain wallet
+      # crate must stay free of LDK so it remains independently replaceable.
+      - name: Ensure the on-chain wallet stays free of LDK
+        run: ./tools/check-wallet-decoupling.sh
--- a/docs/designs/unified-chain-sync.md
+++ b/docs/designs/unified-chain-sync.md
@@ -1,17 +1,39 @@
-# Unified Chain Sync: Align Lampo with ldk-node
+# Unified Chain Sync: Align Lampo with ldk-node *without losing backend/wallet pluggability*
 
 | Field | Value |
 |-------|--------|
-| Status | Draft |
-| Author | Vincenzo Palazzo (with Grok-assisted analysis) |
+| Status | Draft (rev. 2 — decoupled) |
+| Author | Vincenzo Palazzo (with Grok-assisted analysis; rev. 2 decoupling pass) |
 | Date | 2026-06-03 |
-| Related issues | [#540](https://github.com/vincenzopalazzo/lampo.rs/issues/540) (fast sync), [#444](https://github.com/vincenzopalazzo/lampo.rs/issues/444) (`reindex` semantics) |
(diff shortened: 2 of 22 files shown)


vincenzopalazzo:
Closing this in favor of #551. While consolidating I found that this fix was actually a no-op :/ the call goes through the Backend trait's default empty set_coordinator, so it never reaches the real backend (LampoChainManager does not override it). The real bug observed live was the missing RPC timeout, fixed in the new series, and the facade now forwards all the injection hooks so this foot-gun is gone.

PR #549, opened 2026-07-06
Fix(sync): add wallet-gate escape path and bound the listener sync

vincenzopalazzo:
## Summary

Fixes the second half of the PR #546 chain-sync deadlock: a hung or very slow backend permanently blocks the on-chain wallet.

- **What was broken:** `synchronize_listeners` uses `lightning_block_sync::rpc::RpcClient`, which has **no read timeout**. On a degraded backend (e.g. the SSH-tunneled signet node) an RPC call never returns, so `synchronize_listeners` hangs forever. Because the wallet cron is **hard-gated** behind `chain_listeners_synced` with no escape path, the wallet makes **zero progress** indefinitely — whereas `main` scans the wallet in parallel and progresses.
- **Root cause:** (1) no escape path in the wallet gate; (2) no bound on a single `synchronize_listeners` pass.
- **Fix:**
  1. **Escape path** — `ChainSyncCoordinator::wallet_scan_allowed()` returns `true` once listeners are synced *or* after a grace window (`LISTENER_SYNC_GRACE` = 300 s). Past the grace the wallet scans independently, mirroring `main`, so a hung/slow backend can't permanently block it. Logged once (guarded by an `AtomicBool`).
  2. **Bound the listener sync** — wrap each `synchronize_listeners` call in a `tokio::time::timeout` (120 s). A hung RPC now times out and the existing retry loop resumes from the listeners' current best block instead of stalling forever.

This complements #548 (which makes `mark_listeners_synced` reachable by wiring the coordinator into the on-chain manager). Both are needed for PR #546 to function end-to-end on a wallet with a backlog.

## Test plan

- [x] `cargo check --workspace --all-targets` clean (only pre-existing warnings)
- [x] `cargo test -p lampo-common chainsync` — 6 tests pass, incl. new `wallet_scan_allowed_reflects_sync_state`
- [x] **Verified live on signet** (restored wallet, height 21021 vs tip 312046):
  - Wallet gated for the first ~360 s (`Waiting for chain listeners to sync…`), `getblock` RPC count = 0.
  - At the grace boundary the escape fired once: `listener sync not complete after 360.31s; allowing the on-chain wallet to scan independently`.
  - The wallet cron immediately transitioned from `Waiting…` to `Tick tock… Checking the wallet status` and began scanning.
  - `getblock` RPC count climbed 0 → 85 → 208 → 301 (~3.7 blocks/s) — i.e. the node now progresses exactly like `main`, instead of being permanently deadlocked.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -88,3 +88,8 @@ jobs:
             exit 1
           fi
           echo "OK: each lightning* crate resolves to a single version."
+
+      # Guards the unified-chain-sync invariant (#545): the on-chain wallet
+      # crate must stay free of LDK so it remains independently replaceable.
+      - name: Ensure the on-chain wallet stays free of LDK
+        run: ./tools/check-wallet-decoupling.sh
--- a/docs/designs/unified-chain-sync.md
+++ b/docs/designs/unified-chain-sync.md
@@ -1,17 +1,39 @@
-# Unified Chain Sync: Align Lampo with ldk-node
+# Unified Chain Sync: Align Lampo with ldk-node *without losing backend/wallet pluggability*
 
 | Field | Value |
 |-------|--------|
-| Status | Draft |
-| Author | Vincenzo Palazzo (with Grok-assisted analysis) |
+| Status | Draft (rev. 2 — decoupled) |
+| Author | Vincenzo Palazzo (with Grok-assisted analysis; rev. 2 decoupling pass) |
 | Date | 2026-06-03 |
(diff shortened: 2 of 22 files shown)


vincenzopalazzo:
Closing this in favor of #551. Both fixes (the 120s bound on synchronize_listeners and the 300s wallet-gate escape path) are folded into the new series, adapted to the retry loop with exponential backoff that #546 grew after review.

PR #552, opened 2026-08-09
Close on-chain fund-loss holes for production use

vincenzopalazzo:
## Summary

Make lampo safe to run on mainnet with real funds by closing every on-chain fund-loss hole found in a full audit of the tree. One commit per concern, each self-contained; the full audit, design decisions, and justification live in [`docs/designs/production-onchain-safety.md`](docs/designs/production-onchain-safety.md) (added by the last commit, together with the brainstorm that ranked the risks).

### Fund-loss holes closed

1. **Channel monitors were never registered on restart** — no `watch_channel` call existed anywhere in the tree. After any restart the `ChainMonitor` was empty: counterparty force-closes, revoked commitments, and on-chain HTLCs went undetected for every existing channel. Monitors are now synced each from its own persisted best block and registered before the node goes live; startup aborts if registration fails.
2. **`Event::SpendableOutputs` was dropped** — force-close outputs were never claimed. LDK's `OutputSweeper` is now wired end to end (persisted state, BDK change addresses, chain listener, background-processor timer). Persist failures replay the event instead of dropping it.
3. **`FeeEstimator` returned a hardcoded 256 sat/kW** for every target, and the funding path overpaid fees ~1000x from a sat/kvB-as-sat/vB unit mix-up. Feerates now come from a per-target cache refreshed from `estimatesmartfee`/`getmempoolinfo`, unified on sat-per-1000-weight, with headroom on `MaximumFeeEstimate`.
4. **No `BumpTransactionEventHandler`** despite anchors being negotiated by default — anchor commitments *rely* on CPFP to confirm. The handler is wired with coin selection from the BDK wallet; inbound anchor channels now require a 25k-sat confirmed reserve.
5. **Broadcast failures were silently swallowed**, and multi-tx broadcasts broke LDK's package contract. `brodcast_tx` returns `Result`, packages go through bitcoind's `submitpackage`, "already known" counts as success, bounded retries, and definitive failures emit a `BroadcastFailed` event so RPC callers stop hanging.
6. **Startup ordering**: pid lock now taken before any state is read; peers/events/RPC start only after the initial chain sync (and monitor registration) completes; `PaymentClaimable` no longer `unwrap()`s the preimage; corrupt graph/scorer files warn loudly; `accept-inbound-channels` config option added.

### Design

Follows the plan in `docs/brainstorms/2026-08-09-lampo-production-fund-safety.md` (Approach A: onchain safety first, reusing LDK components — `OutputSweeper`, `BumpTransactionEventHandler`, `Wallet`/`WalletSource` — behind thin `WalletManager` adapters, keeping lampo's facade/backend architecture). Two items of `docs/designs/unified-chain-sync.md` (per-listener locators, monitor registration) land here ahead of the coordinator work; the justification for that ordering is in the new design doc, and the unified-chain-sync doc carries a status note. VSS channel-state mirroring and the sync coordinator are the next phases and deliberately out of scope.

### Testing

- `cargo test --all`: unit tests pass; the `tests` crate shows pre-existing environment flakiness (`[43] libcurl` errors) that reproduces identically on `main` at 4a5afd8.
- Every commit compiles independently (`cargo check --workspace` was run per-commit during the rebase).
- An adversarial review pass against the pinned LDK 0.3 sources verified the sweeper `ReadableArgs` ordering, the restart flow against LDK's canonical init example, fee unit conversions, and absence of lock-across-await; the issues it found (package broadcast, blanket replay, missing anchor reserve, early RPC exposure) are fixed in this series.

### Follow-ups (tracked in the design doc)

VSS mirror for channel state, unified chain sync coordinator, anchor reserve *maintenance* (accept-time check only today), payment persistence, mnemonic encryption, and a regtest force-close-and-sweep integration test.

### Fund-loss reproduction (A/B on the new regression test)

`tests: Cover sweeping funds after channel close` opens a 1M-sat channel, cooperatively closes it, and waits for the closed channel's funds to reappear as a wallet UTXO.

- **On `main` (4a5afd8, with only the `/close` endpoint fix applied so the close can happen at all): FAILS in 165s.** The node logs the exact loss — LDK hands over the funds and lampo drops them:
  ```
  WARN lampo::handler unhandled ldk event: SpendableOutputs {
      outputs: [StaticOutput { value: 998828 SAT, ... }], ... }
  ```
  The 998,828 sats stay on the LDK-derived shutdown script forever; no component in the tree can ever spend them.
- **On this branch: PASSES in 91s.** The same event is tracked (`tracking 1 spendable output(s) ...`), the sweeper broadcasts (`Generating and broadcasting sweeping transaction 57860f...`), and the funds land back in the BDK wallet.

Writing this test also flushed out one more pre-existing bug, fixed here: the httpd `/close` endpoint discarded its request body (the body-less `post!` variant), so closing a channel through the HTTP API had never worked at all.

### DIFF
--- a/docs/brainstorms/2026-08-09-lampo-production-fund-safety.md
+++ b/docs/brainstorms/2026-08-09-lampo-production-fund-safety.md
@@ -1,0 +1,151 @@
+# Lampo Production Fund-Safety
+
+Date: 2026-08-09
+Status: Brainstorm (clarified problem statement)
+
+## Clarified Problem Statement
+
+**Goal:** Make Lampo safe to run on mainnet with small funds by closing the
+onchain fund-loss holes (sweeping, fees, broadcast, bumping), adding a
+VSS-mirrored channel-state backup, unifying chain sync per the existing design
+doc, and hardening startup/acceptance paths — reusing LDK's battle-tested
+components (like ldk-node does) while keeping Lampo's own architecture.
+
+**Constraints:**
+- Reuse LDK off-the-shelf components: `lightning::util::sweep::OutputSweeper`,
+  `BumpTransactionEventHandler`, `vss-client` — no Lampo-native rewrites.
+- Filesystem stays the primary store (fast, fsync'd); VSS is a mirror/backup,
+  not the primary KVStore. Recovery from VSS is an explicit tool/flow.
+- Keep Lampo's design solid: small, self-contained PRs per CLAUDE.md; no
+  dependency additions without maintainer sign-off (vss-client is the main new one).
+- bitcoind RPC remains the only chain backend for now.
+- LDK is pinned at 0.3 (git rev `84605cf`); all work targets that API.
+
+**Non-goals:**
+- Esplora/electrum/compact-filter backends.
+- VSS as primary store (ldk-node's "VSS mode") — mirror only.
+- Adopting ldk-node itself as a dependency; we mirror its patterns, not its crate.
+- Watchtowers, splicing, LSP flows.
+- Mnemonic encryption (worth doing, tracked separately — not fund-loss via LN).
+
+**Success criteria:**
+- Integration test: force-close a channel on regtest → `SpendableOutputs` is
+  swept back into the BDK wallet automatically (sweeper survives restart mid-sweep).
+- `FeeEstimator` serves per-`ConfirmationTarget` rates from `estimatesmartfee`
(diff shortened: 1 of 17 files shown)


vincenzopalazzo:
Review fix pushed in b61eda2: WalletSource now rejects partially finalized PSBTs instead of extracting and broadcasting them. cargo check --workspace --all-targets passed locally and the fix-delta panel found no regression. Package broadcast semantics, fee-cache staleness, and generic backend no-op hooks still require explicit design work before production readiness.

PR #554, opened 2026-08-09
Move LDK 0.3 to the crates.io release, add BOLT 12 payer proofs

vincenzopalazzo:
Two commits, reviewable independently. The second depends on the first: the `payer_proof` module does not exist at the rev we were pinned to.

## `build: move LDK 0.3 to the crates.io release`

- LDK 0.3 is now published as `0.3.0-beta1`, so drop the git pin on the upstream `0.3` branch and pull every `lightning*` crate from the registry
- The published beta is **96 commits ahead** of the rev we were pinned to (`84605cf`)
- The requirement spells out `0.3.0-beta1` on purpose -- a bare `0.3` does not match pre-release versions
- `Cargo.lock` still pins exact versions, so builds stay reproducible without the git source
- `deny.toml` sets `allow-git = []`; removing the last git dep drops that exception

No API drift. `lightning-macros` moves `0.2.2+git` -> `0.2.1`, which is just the released equivalent of the in-tree version.

## `node: attach BOLT 12 payer proof to pay response`

`pay` on an offer now returns a bech32 `payer_proof` plus the `payment_preimage`, so the payer can prove to a third party that they paid that invoice.

- `Event::PaymentSent` was being discarded. It is the only event carrying the paid invoice, so the handler now stores that invoice plus the preimage under the payment id.
- Only those two things are stored. The payer signing key is re-derived from the invoice's own payer metadata, so no nonce is needed, and the payment id is recoverable from the same metadata (kept only as the lookup key).
- We store rather than build-and-forget because disclosure is committed to at signing time: widening a proof later means rebuilding it from the original invoice.
- Disclosure defaults to LDK's minimal set. A proof can be widened later but never narrowed once handed out.
- `PaymentPathSuccessful` remains the event that ends the pay call, so `PayResult.path` is unchanged. Returning on `PaymentSent` instead would have silently emptied it -- that event carries no hop data.
- Closes the `FIXME: missing payment preimage` on `PayResult`.
- BOLT 11 and async (static invoice) payments cannot produce a proof and report `payer_proof: null` rather than failing.

Design notes in `docs/brainstorms/2026-08-09-attach-payer-proof-to-pay-response.md`.

Known gap, marked FIXME: stored records are never pruned, so a node keeps one invoice per BOLT 12 payment indefinitely. Wants a retention policy or a drop RPC before this is load-bearing.

## Test plan

- `cargo check --workspace --all-targets`, `cargo test`, `cargo build --release`, `cargo fmt --all -- --check` -- all clean
- Single-LDK-version guard from `build.yml` run locally
- 3 unit tests on the stored record: roundtrip, truncated input, unknown version
- `pay_offer_simple_case_lampo` now parses the returned proof and asserts its `payment_hash` and `payment_preimage` match what the RPC reported -- this exercises the whole chain, LDK really produces a valid proof
- `pay_invoice_simple_case_lampo` asserts the BOLT 11 case yields a preimage but no proof
- Payment/channel integration tests run individually: all pass

Note the integration job is flaky under parallel load, unrelated to either commit -- see the earlier comment on this PR.

### DIFF
--- a/lampod/src/actions/handler.rs
+++ b/lampod/src/actions/handler.rs
@@ -17,12 +17,15 @@ use lampo_common::handler::Handler as EventHandler;
 use lampo_common::json;
 use lampo_common::jsonrpc::Request;
 use lampo_common::ldk;
+use lampo_common::ldk::sign::NodeSigner;
 use lampo_common::model::response::PaymentHop;
 use lampo_common::model::response::PaymentState;
 
 use crate::chain::{LampoChainManager, WalletManager};
 use crate::command::Command;
+use crate::ln::payer_proof::{self, PayerProofRecord};
 use crate::ln::{LampoChannelManager, LampoInventoryManager, LampoPeerManager};
+use crate::persistence::LampoPersistence;
 use crate::LampoDaemon;
 
 use super::Handler;
@@ -33,6 +36,7 @@ pub struct LampoHandler {
     inventory_manager: Arc<LampoInventoryManager>,
     wallet_manager: Arc<dyn WalletManager>,
     chain_manager: Arc<LampoChainManager>,
+    persister: Arc<LampoPersistence>,
     external_handlers: RwLock<Vec<Arc<dyn ExternalHandler>>>,
     #[allow(dead_code)]
     emitter: Emitter<Event>,
@@ -49,6 +53,7 @@ impl LampoHandler {
             inventory_manager: lampod.inventory_manager(),
             wallet_manager: lampod.wallet_manager(),
             chain_manager: lampod.onchain_manager(),
+            persister: lampod.persister(),
             external_handlers: RwLock::new(Vec::new()),
             emitter,
             subscriber,
@@ -378,12 +383,64 @@ impl Handler for LampoHandler {
                 // FIXME: make peristent these information
                 Ok(())
             }
-            ldk::events::Event::PaymentSent { .. } => {
-                log::info!("payment sent: `{:?}`", event);
+            ldk::events::Event::PaymentSent {
+                payment_id,
+                payment_preimage,
+                payment_hash,
+                bolt12_invoice,
+                ..
+            } => {
+                log::info!(
+                    target: "lampo::handler",
+                    "payment sent: id `{payment_id:?}`, bolt12 payer proof available: {}",
+                    bolt12_invoice.is_some()
+                );
+                let Some(payment_id) = payment_id else {
+                    // Without an id the receipt cannot be matched to the `pay`
+                    // call that is waiting for it, so there is nothing to emit.
+                    return Ok(());
+                };
+                let record = PayerProofRecord {
+                    preimage: payment_preimage,
+                    invoice: bolt12_invoice,
+                };
+
+                // Build the proof from the material in hand rather than from
+                // storage, so the receipt does not depend on the write below.
+                let expanded_key = self
+                    .wallet_manager
+                    .ldk_keys()
+                    .keys_manager
+                    .get_expanded_key();
+                let payer_proof = payer_proof::build(&record, &expanded_key, payment_id);
+
+                self.emit(Event::Lightning(LightningEvent::PaymentReceipt {
+                    payment_id: lampo_common::hex::encode(payment_id.0),
+                    payment_preimage: lampo_common::hex::encode(record.preimage.0),
+                    payer_proof,
+                }));
+
+                // This is the only event carrying the paid BOLT 12 invoice, and the
+                // proof cannot be rebuilt without it. Keep it so the proof can be
+                // re-issued later with a wider disclosure. The payment already
+                // settled and the receipt is already out, so a storage failure
+                // only costs us the ability to re-issue.
+                //
+                // Only BOLT 12 payments are worth keeping: a BOLT 11 record holds
+                // a preimage and nothing else, can never build a proof, and
+                // nothing prunes it.
+                if record.invoice.is_some() {
+                    if let Err(err) = payer_proof::store(&self.persister, &payment_hash, &record) {
+                        log::error!(target: "lampo::handler", "storing payer proof material: {err}");
+                    }
+                }
                 Ok(())
             }
             ldk::events::Event::PaymentPathSuccessful {
-                payment_hash, path, ..
+                payment_id,
+                payment_hash,
+                path,
+                ..
             } => {
                 let path = path
                     .hops
@@ -392,6 +449,7 @@ impl Handler for LampoHandler {
                     .collect::<Vec<PaymentHop>>();
                 let hop = LightningEvent::PaymentEvent {
                     state: PaymentState::Success,
+                    payment_id: Some(lampo_common::hex::encode(payment_id.0)),
                     payment_hash: payment_hash.map(|hash| hash.to_string()),
                     path,
                     reason: None,
@@ -447,6 +505,7 @@ impl Handler for LampoHandler {
 
                 let hop = LightningEvent::PaymentEvent {
                     state: PaymentState::Failure,
+                    payment_id: Some(lampo_common::hex::encode(payment_id.0)),
                     payment_hash: payment_hash.map(|hash| hash.to_string()),
                     path: vec![],
                     reason: Some(detailed_reason),
--- a/lampod/src/jsonrpc/offchain.rs
+++ b/lampod/src/jsonrpc/offchain.rs
@@ -4,6 +4,7 @@ use std::str::FromStr;
 use lampo_common::event::ln::LightningEvent;
 use lampo_common::event::Event;
 use lampo_common::handler::Handler;
+use lampo_common::hex;
 use lampo_common::jsonrpc::{Error, RpcError};
 use lampo_common::ldk;
 use lampo_common::ldk::offers::offer;
@@ -98,19 +99,28 @@ pub async fn json_pay(ctx: &LampoDaemon, request: &json::Value) -> Result<json::
     let request: Pay = json::from_value(request.clone())?;
     let mut events = ctx.handler().events();
 
-    if let Ok(_) = offer::Offer::from_str(&request.invoice_str) {
+    let payment_id = if let Ok(_) = offer::Offer::from_str(&request.invoice_str) {
         log::debug!("Paying offer with bolt12 invoice: {}", request.invoice_str);
         let payer_note = request.bolt12.and_then(|x| x.payer_note);
         ctx.offchain_manager()
-            .pay_offer(&request.invoice_str, request.amount, payer_note)?;
+            .pay_offer(&request.invoice_str, request.amount, payer_note)?
     } else {
         log::debug!(
             "Paying invoice with bolt11 invoice: {}",
             request.invoice_str
         );
         ctx.offchain_manager()
-            .pay_invoice(&request.invoice_str, request.amount)?;
-    }
+            .pay_invoice(&request.invoice_str, request.amount)?
+    };
+    // The event bus broadcasts to every subscriber, so a concurrent `pay` would
+    // otherwise see this payment's result -- and now its preimage and payer
+    // proof too. Only accept events carrying our own payment id.
+    let payment_id = hex::encode(payment_id.0);
+
+    // The receipt lands on `PaymentReceipt` and the hop path on the terminal
+    // `PaymentEvent`, so hold the receipt until the payment finishes.
+    let mut receipt: Option<(String, Option<String>)> = None;
+
     // FIXME: this will loop when the Payment event is not generated
     loop {
         log::warn!("Waiting for payment event...");
@@ -120,18 +130,34 @@ pub async fn json_pay(ctx: &LampoDaemon, request: &json::Value) -> Result<json::
             data: None,
         }))?;
 
-        if let Event::Lightning(LightningEvent::PaymentEvent {
-            payment_hash,
-            path,
-            state,
-            reason: _,
-        }) = event
-        {
-            return Ok(json::to_value(PayResult {
-                state,
-                path,
+        match event {
+            Event::Lightning(LightningEvent::PaymentReceipt {
+                payment_id: id,
+                payment_preimage,
+                payer_proof,
+            }) if id == payment_id => {
+                receipt = Some((payment_preimage, payer_proof));
+            }
+            Event::Lightning(LightningEvent::PaymentEvent {
+                payment_id: Some(id),
                 payment_hash,
-            })?);
+                path,
+                state,
+                reason: _,
+            }) if id == payment_id => {
+                let (payment_preimage, payer_proof) = match receipt {
+                    Some((preimage, proof)) => (Some(preimage), proof),
+                    None => (None, None),
+                };
+                return Ok(json::to_value(PayResult {
+                    state,
+                    path,
+                    payment_hash,
+                    payment_preimage,
+                    payer_proof,
+                })?);
+            }
+            _ => {}
         }
     }
 }
--- a/lampod/src/lib.rs
+++ b/lampod/src/lib.rs
@@ -134,6 +134,10 @@ impl LampoDaemon {
         self.conf.clone()
     }
 
+    pub fn persister(&self) -> Arc<LampoPersistence> {
+        self.persister.clone()
+    }
+
     pub fn init_onchaind(&mut self, client: Arc<dyn Backend>) -> error::Result<()> {
         log::debug!(target: "lampod", "init onchaind ..");
         let onchain_manager = LampoChainManager::new(client, self.wallet_manager.clone());
--- a/lampod/src/ln/payer_proof.rs
+++ b/lampod/src/ln/payer_proof.rs
@@ -1,0 +1,198 @@
+//! BOLT 12 payer proof material.
+//!
+//! LDK hands us the paid invoice on `Event::PaymentSent` and nowhere else, so we
+//! store it together with the preimage. Disclosure is committed to when the proof
+//! is signed, which means a proof cannot be widened after the fact: rebuilding it
+//! with more disclosed fields requires the original invoice.
+//!
+//! Nothing else needs storing. The payer signing key is re-derived from the
+//! invoice's own payer metadata, and the `ExpandedKey` behind that derivation
+//! comes from the node seed.
+use std::sync::Arc;
+
+use lampo_common::error;
+use lampo_common::ldk::events::PaidBolt12Invoice;
+use lampo_common::ldk::io::{Cursor, ErrorKind};
+use lampo_common::ldk::ln::channelmanager::PaymentId;
+use lampo_common::ldk::ln::inbound_payment::ExpandedKey;
+use lampo_common::ldk::types::payment::{PaymentHash, PaymentPreimage};
+use lampo_common::ldk::util::persist::KVStoreSync;
+use lampo_common::ldk::util::ser::{Readable, Writeable};
+use lampo_common::secp256k1::Secp256k1;
+
+use crate::persistence::LampoPersistence;
+
+/// Namespace holding the payer proof material, keyed by hex payment id.
+///
+/// FIXME: records are never pruned, so a node accumulates one invoice per BOLT 12
+/// payment forever. Needs a retention policy, or an RPC to drop a record once the
+/// payer no longer cares about proving that payment.
+pub const PAYER_PROOF_NAMESPACE: &str = "payer_proofs";
+
+/// Layout version of a stored record, so the format can change later without
+/// misreading old entries.
+const RECORD_VERSION: u8 = 1;
+
+/// Byte offsets in an encoded record: version, then preimage, then the flag
+/// saying whether an invoice follows.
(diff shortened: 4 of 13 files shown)


vincenzopalazzo:
CI went green on a re-run -- the integration failure was a flake, not this bump. Writing up what I checked so nobody has to redo it.

The job was killed by `timeout-minutes: 20`, not by an assertion. It hung ~14 min in `lampo_tests::fund_a_simple_channel_from`, and the log for that whole window is nothing but wallet idle lines:

```
DEBUG lampo-wallet Tick tock, time to check if we need to sync the wallet.
INFO  lampo-wallet Checking the wallet status....
DEBUG bitcoincore_rpc JSON-RPC request: getblock ["6f21f513...", 1]
##[error]The operation was canceled.
```

Ran the same parallel suite locally on both configs:

| config | result |
|---|---|
| this branch (`0.3.0-beta1`) | 4/9 failed |
| base commit (old git pin) | 5/9 failed |
| this branch, test in isolation | passes in 4.45s |

So base fails *more* than the branch under the same conditions. The failures are not LN-related either -- they panic on the `funds` RPC `.unwrap()` in `lampo-testing/src/lib.rs:239` with `[43] A libcurl function was given a bad argument` and `Lost connection to the RPC socket`, i.e. concurrent bitcoind RPC contention in the harness.

BTW the reason it hangs instead of failing is `async_wait!` in `lampo-testing/src/lib.rs:41`:

```rust
let result = $callback.await;   // unbounded
if let Err(_) = result {
    tokio::time::sleep(Duration::from_secs($timeout)).await;
    continue;
}
```

`$timeout` is only the sleep *between* retries -- it never bounds the callback. When the `funds` RPC stalls rather than errors, the test blocks forever and we get a timeout kill with no error message instead of a failed assertion. Wrapping the callback in `tokio::time::timeout(...)` would turn these into a normal retry-then-fail and make the flakes diagnosable. Valid improvement for a follow-up, out of scope here -- this PR is a dependency bump and touches no test code.

lampod/src/actions/handler.rs
```diff
@@ -49,12 +53,47 @@ impl LampoHandler {
...
+    ///
+    /// Both are best effort: a payment that cannot produce a proof (BOLT 11, or
+    /// an async payment settled against a static invoice) still succeeded, so a
+    /// miss here is logged rather than propagated.
+    ///
+    /// This reads what `Event::PaymentSent` stored, so it only returns something
+    /// once that event has been handled. LDK queues `PaymentSent` first, but a
+    /// multipath payment calls this once per path, re-signing the proof each
+    /// time. FIXME: cache the built proof if that ever shows up in a profile.
+    fn payer_proof_for(
```
vincenzopalazzo: maybe thsi is better to have inside the pay/manager handler?

lampod/src/ln/payer_proof.rs
```diff
@@ -0,0 +1,182 @@
...
+        })
+    }
+}
+
+fn key(payment_id: &PaymentId) -> String {
+    lampo_common::hex::encode(payment_id.0)
+}
+
+/// Persist the proof material for `payment_id`, overwriting any earlier record.
+pub fn store(
```
vincenzopalazzo, line 111: can be stored inside the persister interface? otherwise if we cange the persister implementation this code will need some refactorying

lampod/src/ln/payer_proof.rs
```diff
@@ -0,0 +1,182 @@
...
+    persister: &Arc<LampoPersistence>,
+    payment_id: &PaymentId,
+    record: &PayerProofRecord,
+) -> error::Result<()> {
+    persister.write(PAYER_PROOF_NAMESPACE, "", &key(payment_id), record.encode())?;
+    Ok(())
+}
+
+/// Load the proof material for `payment_id`, if any was stored.
+pub fn load(
```
vincenzopalazzo, line 126: same for

lampod/src/lib.rs
```diff
@@ -134,6 +134,10 @@ impl LampoDaemon {
         self.conf.clone()
     }
 
+    pub fn persister(&self) -> Arc<LampoPersistence> {
```
vincenzopalazzo, line 137: `LampoPersistence` shoudl containes the interface to allow us to switch the persistence without having breaking change in lampo, so we need to rework on the trait design

tests/tests/src/lampo_tests.rs
```diff
@@ -181,6 +195,28 @@ pub async fn pay_offer_simple_case_lampo() -> error::Result<()> {
...
+    assert_eq!(
+        proof.payment_hash().to_string(),
+        pay.payment_hash.unwrap(),
+        "the proof must commit to the hash of the payment we made"
+    );
+    assert_eq!(
+        hex::encode(proof.payment_preimage().0),
+        preimage,
+        "the proof must carry the same preimage the RPC returned"
+    );
```
vincenzopalazzo, line 224: we should be able to call verify on top of the payer proof

vincenzopalazzo (commented):
some comments

lampod/src/jsonrpc/offchain.rs
```diff
@@ -125,12 +125,16 @@ pub async fn json_pay(ctx: &LampoDaemon, request: &json::Value) -> Result<json::
...
+            payment_preimage,
+            payer_proof,
         }) = event
         {
             return Ok(json::to_value(PayResult {
                 state,
                 path,
                 payment_hash,
+                payment_preimage,
+                payer_proof,
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in 8eec791.** Good catch -- and it was worse than the pre-existing version, since the leaked fields now include the preimage and the payer proof.

`pay_offer` and `pay_invoice` now return the `PaymentId` they used (both are deterministic already -- `Sha256(offer_str)` and the invoice payment hash), every payment event carries a hex `payment_id`, and `json_pay` accepts only events matching its own. Filtering on the id rather than deriving the hash also works for offers, where the payment hash is not known until the invoice comes back.

lampod/src/actions/handler.rs
```diff
@@ -378,23 +417,52 @@ impl Handler for LampoHandler {
...
+                if let Some(payment_id) = payment_id {
+                    let record = PayerProofRecord {
+                        preimage: payment_preimage,
+                        invoice: bolt12_invoice,
+                    };
+                    // The payment already settled, so a storage failure must not
+                    // fail the event: it only costs us the proof.
+                    if let Err(err) = payer_proof::store(&self.persister, &payment_id, &record) {
+                        log::error!(target: "lampo::handler", "storing payer proof material: {err}");
+                    }
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in 8eec791.** The receipt no longer round-trips through storage at all.

`PaymentSent` now builds the proof from the material in hand and emits a `PaymentReceipt` event immediately; the persister write happens after, purely so the proof can be re-issued later. A failed write costs the re-issue path, not the current RPC response. This also drops the per-path disk read the old version did on every `PaymentPathSuccessful`.

lampod/src/actions/handler.rs
vincenzopalazzo: **Fixed in 8eec791.** The helper is gone entirely.

It only existed to read the material back from disk on `PaymentPathSuccessful`. Now `PaymentSent` builds the proof where the material arrives and emits it, so there is nothing left to look up -- the handler just calls into `ln/payer_proof.rs` and the reusable logic lives there rather than in the handler.

I kept the call in the event handler rather than moving it to `OffchainManager` because `PaymentSent` is an LDK event and `LampoHandler` is what receives those; the handler would have to reach into the offchain manager to hand the data back. Happy to move it if you would rather the manager own the whole payment lifecycle, but that felt like a bigger restructure than this PR.

lampod/src/ln/payer_proof.rs
vincenzopalazzo: **Fixed in 8eec791.** Added a `PayerProofStore` trait with the store/load pair, and a blanket impl over `KVStoreSync`:

```rust
impl<T> PayerProofStore for T where T: KVStoreSync + ?Sized { ... }
```

So the payment code never names a concrete store, and any `KVStore`-backed persister we switch to keeps working without touching this file.

lampod/src/ln/payer_proof.rs
vincenzopalazzo: Same commit as above -- `load` moved onto the same `PayerProofStore` trait.

lampod/src/lib.rs
vincenzopalazzo: Agreed on the direction, but I did the scoped half here and want to leave the rest separate.

What is in 8eec791: payer proof storage goes through a `PayerProofStore` trait with a blanket impl over `KVStoreSync`, so *this* code path is already backend-agnostic -- swapping the store needs no change in the payment code.

What I did not do: turn `LampoPersistence` itself from a `type LampoPersistence = FilesystemStore` alias into a real trait. That one alias is threaded through the channel manager, chain monitor and `read_channel_monitors`, all of which are typed against LDK traits that want a concrete `KVStoreSync`. Reworking it is a real design job and would swamp a PR that is meant to be an LDK bump plus one feature.

Happy to open a follow-up for the `LampoPersistence` trait rework, or do it here if you would rather it land together -- let me know.

tests/tests/src/lampo_tests.rs
vincenzopalazzo: There is no separate verify entry point to call -- LDK verifies while parsing. `PayerProof::from_str` goes through `TryFrom<Vec<u8>>`, which checks the preimage against the payment hash, the invoice signature against the issuer key, and the payer signature over the merkle root (`payer_proof.rs:902-921`). So the test was already verifying, it just did not say so.

**Made that explicit in 8eec791**: comment naming the three checks, and a negative case that mutates a character of the proof payload and asserts it no longer parses -- so the test fails if verification ever silently degrades into a plain decode.

lampod/src/lib.rs
vincenzopalazzo: Done in 553ca07 -- you were right and my earlier reply was too confident.

I said the blanket impl over `KVStoreSync` gave us swappability. It does not, and Postgres is exactly the case that breaks it: LDK needs `KVStoreSync` for channel monitors, so a Postgres backend implements it too, so the blanket impl claims it -- and coherence then forbids that backend from writing its own native impl. I checked with a reduced case, it is a hard E0119:

```
error[E0119]: conflicting implementations of trait `PayerProofStore` for type `PostgresStore`
```

You would be stuck writing an opaque blob into a KV-emulation table.

So `LampoPersistence` is now a trait, held as `Arc<dyn LampoPersistence>`, and the domain methods take the object rather than bytes -- `store_payer_proof(&self, payment_id, &PayerProofRecord)`. The KV impl encodes a versioned blob; a Postgres impl writes columns and does whatever it likes. No blanket impl, so each backend is free.

Two things that made it work, in case they bite you elsewhere:
- `KVStoreSync` stays a supertrait, so `Arc<dyn LampoPersistence>` still satisfies LDK. Their `impl<ChannelSigner, K: KVStoreSync + ?Sized> Persist<ChannelSigner> for K` is `?Sized`, which is what lets the trait object through.
- The async `KVStore` trait is RPITIT and therefore not object safe, so `process_events_async` takes the store via LDK own `KVStoreSyncWrapper`.

One thing I left sync: the trait methods. A `tokio-postgres` backend would need `block_on`, or we make the two domain methods async with `async_trait` -- two call sites, cheap either way. IDK which you prefer, happy to flip it now.

vincenzopalazzo: Update: I have pulled the persistence work out of this PR entirely, per your call that it wants its own.

553ca07 is gone from this branch. What is left here is the LDK bump plus the payer proof feature, and payer proof material is stored through two free functions taking the concrete store -- no trait. That is deliberate rather than lazy: the interface changed shape three times in one sitting (blanket impl -> trait -> async mirror), and every redesign came from reasoning about a backend that does not exist yet. An interface with one implementation is how you get the wrong interface.

The persistence work moves to a single PR doing it async-first, on a branch stacked on this one. Two things I found while prototyping it that are worth knowing before that review:

1. The monitor migration I was worried about is a non-issue. `MonitorUpdatingPersister` strips its sentinel only if present, so existing plain-format monitors deserialize fine and new writes upgrade in place. Downgrade is the one-way direction.
2. LDK labels the async monitor path beta. `ChainMonitor::new_async_beta` is the only async constructor, and its own docs say "async monitor updating is considered beta, and bugs may be triggered by its use." Going async-first means running that on channel monitors. Worth deciding with eyes open -- the sync `MonitorUpdatingPersister` gets the identical on-disk format without the beta constructor, if we ever want to split the difference.

Ill open the persistence PR against this branch so the diff stays clean.

lampod/src/ln/payer_proof.rs
```diff
@@ -0,0 +1,187 @@
...
+    persister: &Arc<LampoPersistence>,
+    payment_id: &PaymentId,
+    record: &PayerProofRecord,
+) -> error::Result<()> {
+    persister.write(PAYER_PROOF_NAMESPACE, "", &key(payment_id), record.encode())?;
+    Ok(())
+}
+
+/// Load the material for `payment_id`, if any was stored.
+pub fn load(
```
vincenzopalazzo, line 126: we should refactor this in a separate payment store struct and probably this need to be under the `LampoStorageLayer`

tests/tests/src/lampo_tests.rs
```diff
@@ -181,6 +195,55 @@ pub async fn pay_offer_simple_case_lampo() -> error::Result<()> {
...
+        .map(|(i, c)| {
+            if i == 0 && c != 'q' {
+                'q'
+            } else if i == 0 {
+                'p'
+            } else {
+                c
+            }
+        })
+        .collect();
```
vincenzopalazzo: Mh this is super strange. we do not have a `PayerProof::verify` or something? that just make sure that the payer proof is valid?

lampod/src/actions/handler.rs
```diff
@@ -378,12 +384,57 @@ impl Handler for LampoHandler {
...
+                    payment_preimage: lampo_common::hex::encode(record.preimage.0),
+                    payer_proof,
+                }));
+
+                // This is the only event carrying the paid BOLT 12 invoice, and the
+                // proof cannot be rebuilt without it. Keep it so the proof can be
+                // re-issued later with a wider disclosure. The payment already
+                // settled and the receipt is already out, so a storage failure
+                // only costs us the ability to re-issue.
+                if let Err(err) = self.persister.store_payer_proof(&payment_id, &record) {
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in 335a7da.** Good catch -- the write is now gated on `record.invoice.is_some()`, so a BOLT 11 payment stores nothing.

You are right that the record was useless: without a `PaidBolt12Invoice` there is nothing `prove_payer_derived` can be called on, so it could only ever hold a preimage that no one reads. Combined with no pruning, a BOLT 11-heavy node was accumulating one dead file per payment forever.

lampo-common/src/persistence.rs
```diff
@@ -0,0 +1,172 @@
...
+impl LampoPersistence for FilesystemStore {
+    fn store_payer_proof(
+        &self,
+        payment_id: &PaymentId,
+        record: &PayerProofRecord,
+    ) -> error::Result<()> {
+        self.write(
+            PAYER_PROOF_NAMESPACE,
+            "",
+            &payer_proof_key(payment_id),
```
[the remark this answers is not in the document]
vincenzopalazzo: **Fixed in 335a7da.** Records are now keyed by `payment_hash` rather than `payment_id`.

Worth spelling out what I checked, because the window is narrower than it first looks but the bug is real. `pay_for_offer` rejects a *concurrent* reuse -- `add_new_awaiting_invoice` returns `Bolt12SemanticError::DuplicatePaymentId` while the first payment is still pending. But once that payment completes and LDK prunes it, the id is free again, and the second payment of the same reusable offer then lands on the same key and overwrites the first record. So it is delayed, not prevented.

`payment_hash` is unique per payment regardless, and it is also what the caller already has from the `pay` response, so it is the better lookup key anyway.

Note the offer-derived `PaymentId` itself is pre-existing (`b256307`, "lampod: be able to pay an offer") and I have not changed it -- two payments of one reusable offer still share an id everywhere else in the node. That is worth its own fix, but it is not something this PR introduced.

lampod/src/ln/payer_proof.rs
vincenzopalazzo: Agreed, and that is exactly the persistence PR. These two free functions are the deliberately-dumb placeholder until it lands -- `LampoStorageLayer` (or whatever we name the trait) owns a payment store struct, and payer proofs become one of its methods rather than free functions reaching into the persister.

I did not do it here because the shape kept moving while it had one implementation, which is what you called out when you said the persistence work wants its own PR. Branch is ready off this one.

tests/tests/src/lampo_tests.rs
vincenzopalazzo: There is no `PayerProof::verify` -- LDK verifies while parsing instead, so `from_str` *is* the check. `TryFrom<Vec<u8>> for PayerProof` (`payer_proof.rs:902-921`) verifies three things before it hands you a value:

1. the preimage hashes to the payment hash
2. the invoice signature against the issuer signing pubkey
3. the payer signature over the proof merkle root

via `merkle::verify_signature` for the latter two. So a `PayerProof` you are holding has already been verified; there is nothing left to call.

I agree it reads oddly in a test, which is why the test now says so explicitly and pins it: it mutates a character of the proof payload and asserts it stops parsing. If verification ever silently degraded into a plain decode, that assertion fails.

Happy to add a thin `verify()` helper upstream if you think the API is confusing -- I hit the same "wait, where is verify?" moment reading it.

lampod/src/jsonrpc/offchain.rs
```diff
@@ -120,18 +130,34 @@ pub async fn json_pay(ctx: &LampoDaemon, request: &json::Value) -> Result<json::
...
+                payment_id: Some(id),
                 payment_hash,
-            })?);
+                path,
+                state,
+                reason: _,
+            }) if id == payment_id => {
+                let (payment_preimage, payer_proof) = match receipt {
+                    Some((preimage, proof)) => (Some(preimage), proof),
+                    None => (None, None),
```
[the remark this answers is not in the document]
vincenzopalazzo: Not a bug -- the premise is inverted. `PaymentSent` is emitted on the *first* claimed HTLC, not after the overall payment settles, so it always precedes any `PaymentPathSuccessful`. LDK asserts this rather than merely happening to do it.

Walking the code in `outbound_payment.rs`:

`claim_htlc` (2275-2305) runs on each fulfilled HTLC:

```rust
if !payment.get().is_fulfilled() {
    pending_events.push_back((events::Event::PaymentSent { .. }));  // 2281
    payment.get_mut().mark_fulfilled();                             // 2289
}
if from_onchain { .. push PaymentPathSuccessful .. }                // 2299
```

So the first path to be claimed pushes `PaymentSent` and only then marks the payment fulfilled. For MPP the later paths see `is_fulfilled() == true` and push no second `PaymentSent`, which is the behaviour you are describing -- but by then the first one is already queued.

The off-chain `PaymentPathSuccessful` comes from `finalize_claims` (2313), which opens with:

```rust
assert!(payment.get().is_fulfilled());
```

`is_fulfilled()` is set nowhere except the line right after `PaymentSent` is pushed. So a `PaymentPathSuccessful` that preceded `PaymentSent` would trip that assert inside LDK. The other emission site is the `from_onchain` branch above, which is also after the `PaymentSent` push in the same function.

Both events land in the same `pending_events` VecDeque and are drained in order by `process_events_async`, and our handler emits to the bus in the order it receives them -- so `PaymentReceipt` is always on the bus before the terminal `PaymentEvent` that `json_pay` returns on.

Concretely for a 3-path MPP: path A claims -> `PaymentSent` (receipt queued) -> paths A/B/C finalize -> three `PaymentPathSuccessful`. `json_pay` stashes the receipt on the first, returns on the second. Never the other way round.

Happy to be shown wrong if you have a path to `finalize_claims` that does not go through `claim_htlc` first -- that assert is the thing I would want to see violated.

tests/tests/src/lampo_tests.rs
vincenzopalazzo: **Fixed in b1a7116.** Dropped the tamper-mutation block from the test after we dug into it.

Two reasons. First, it proved less than its comment claimed: BOLT 12 bech32 is `NoChecksum`, and the flipped character lands in the first TLV record, so the rejection comes from TLV structural parsing -- a decoder with no crypto at all would also reject it. Second, what it did prove is LDK's parser contract, which is owned by LDK's own tests plus the invalid vectors in the spec test suite from bolts#1346. Re-testing a dependency badly and confusingly is worse than not testing it.

What stays: `from_str` (which verifies preimage-vs-hash, the invoice signature, and the payer signature -- a `PayerProof` is verified-by-construction) plus the two binding asserts that tie the proof to *this* payment's hash and preimage. Once your upstream `verify()` lands we can make the first step explicit with `proof.verify()`.

Issue #537, opened 2026-06-01
main does not build: lightning 0.1.5 vs 0.2.2 conflict from lightning-persister 0.2.0


vincenzopalazzo:
## Summary

`main` (at least `e87927f`) fails to compile due to two incompatible versions of the `lightning` crate in the dependency graph.

## Steps to reproduce

```bash
git checkout main
cargo build --release -p lampod-cli -p lampo-cli
```

## Error

```
error[E0277]: the trait bound `FilesystemStore: KVStore` is not satisfied
  --> lampo-common/src/backend.rs:57:42
```

Cargo reports multiple versions of `lightning` in the project:

- `lightning@0.1.5` — used by `lampo-common`, `lightning-block-sync`, `lightning-background-processor`, etc.
- `lightning@0.2.2` — pulled in by `lightning-persister@0.2.0`

## Root cause

Dependabot PR #531 (`5c9e46b`) bumped `lightning-persister` from `0.1.0` to `0.2.0` in `lampo-common/Cargo.toml` without upgrading the rest of the LDK stack. `lightning-persister` 0.2.x depends on `lightning` 0.2.x, while the workspace still pins `lightning = "0.1.5"`.

```bash
cargo tree -p lampo-common -i lightning@0.2.2
# lightning v0.2.2
# └── lightning-persister v0.2.0
#     └── lampo-common

cargo tree -p lampo-common -i lightning@0.1.5
# lightning v0.1.5
# ├── lampo-common
# ├── lightning-background-processor v0.1.0
# ├── lightning-block-sync v0.1.0
# ...
```

## Suggested fixes

**Option A (minimal):** Revert `lightning-persister` to `0.1.0` in `lampo-common/Cargo.toml` and update `Cargo.lock` until a full LDK 0.2 migration is planned.

**Option B (proper):** Upgrade the entire LDK dependency set (`lightning`, `lightning-block-sync`, `lightning-background-processor`, `lightning-net-tokio`, `lightning-rapid-gossip-sync`, `lightning-persister`, etc.) to compatible 0.2.x versions in one change.

## Notes

- Branch `fix/fee-estimation-fallback-244` builds and runs; it predates or avoids this broken dependency state.
- Tag `v24.06` also builds.
- Consider configuring Dependabot to avoid isolated major/minor bumps across the LDK crate family, or add a CI job that runs `cargo build --workspace` on every PR to `main`.

PR #555, opened 2026-08-11
Add TEOS watchtower client

vincenzopalazzo:
## What

Lampo can now act as a watchtower client of an [Eye of Satoshi (rust-teos)](https://github.com/talaia-labs/rust-teos) tower: every revoked counterparty commitment produces a signed justice transaction that is encrypted and delivered to the tower, so a cheating peer is punished even while lampo is offline.

Enable it with two config options (both required together):

```
watchtower-url=http://host:9814
watchtower-id=<tower public key>
```

## How

- **`lampo-watchtower`** (new crate): a `Persist` wrapper around the filesystem store following LDK's watchtower flow — on `commitment_signed` it builds the unsigned justice tx via `counterparty_commitment_txs_from_update`/`build_to_local_justice_tx` and queues it; on `revoke_and_ack` it signs via `sign_to_local_justice_tx` and moves it to a delivery outbox. Both queues are file-backed under `<datadir>/watchtower/`, so states captured while the tower is unreachable are delivered after a restart. A background task registers with the tower and drains the outbox with retry/backoff and re-registration on subscription exhaustion; tower receipts are signature-verified against the configured tower id. The tower is never on the channel hot path: capture never fails a persist, and an unreachable tower does not block payments.
- **`lampod`**: the chain monitor persists through the wrapper; KVStore roles (monitor reads, background processor) keep using the inner store. The justice tx destination is a fresh onchain wallet address derived at startup.

## Why no `teos-common` dependency

The original plan was to reuse `teos-common` as a git dependency, but it pins `lightning 0.1` — which would split the LDK stack in two versions, the exact failure mode the CI single-LDK-version guard defends against (#537) — and drags in tonic/prost/rusqlite plus a `protoc` build requirement for a gRPC surface lampo never touches (the client speaks the tower's public HTTP JSON API). The small protocol surface (locator, appointment serialization, chacha20poly1305 blob encryption, zbase32 message signing) is implemented in `lampo-watchtower::teos` instead, validated against teos-common's own test vectors (the encryption vector matches byte-for-byte).

## Testing

- Unit tests: protocol vectors (encryption, locator, sign/recover, wire JSON) and outbox roundtrips.
- Integration (CLN counterparty):
  - **capture**: with an unreachable tower, two payments leave the revoked state's signed justice tx in the durable outbox while the node keeps operating;
  - **full breach** (auto-skipped when `teosd` is not installed): a real `teosd` receives lampo's appointments; the CLN counterparty's commitment is captured with `dev-sign-last-tx`, revoked by a second payment, and broadcast. The tower decrypts the appointment and responds — the test asserts the appointment flips to `dispute_responded` with the penalty tx paying to lampo's wallet. Passes locally end-to-end.

## Follow-ups (out of scope)

- `teosd` provisioning in CI (e.g. a cached binary or nix package) so the breach test runs there.
- Multi-tower fan-out and tower reputation handling.
- `to_self_delay` in appointments is currently 0 (the monitor does not expose the negotiated value; towers ignore the field today).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,7 +40,7 @@ dependencies = [
  "foldhash",
  "futures-core",
  "h2",
- "http",
+ "http 0.2.12",
  "httparse",
  "httpdate",
  "itoa",
@@ -76,7 +76,7 @@ checksum = "14f8c75c51892f18d9c46150c5ac7beb81c95f78c8b83a634d49f4ca32551fe7"
 dependencies = [
  "bytestring",
  "cfg-if",
- "http",
+ "http 0.2.12",
  "regex",
  "regex-lite",
  "serde",
@@ -190,6 +190,16 @@ name = "adler2"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+
+[[package]]
+name = "aead"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
+dependencies = [
+ "crypto-common",
+ "generic-array",
+]
 
 [[package]]
 name = "ahash"
@@ -314,6 +324,12 @@ dependencies = [
  "quote",
  "syn 2.0.100",
 ]
+
+[[package]]
+name = "atomic-waker"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
 name = "autocfg"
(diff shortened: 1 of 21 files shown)


vincenzopalazzo:
The first integration CI round hit the 20-minute cap: the new watchtower capture test and the pre-existing `fund_a_simple_channel_from_lampo_to_cln` were scheduled concurrently, and two CLN+bitcoind+lampo stacks starve the 2-core runner (same reason most CLN payment tests in this file are commented out). The capture test is now `#[ignore]` — run it locally with `cargo test -p tests -- --ignored`, where it passes in ~45s. The teosd breach test stays enabled since it self-skips in milliseconds when `teosd` is not installed. Both integration tests pass locally end-to-end.

vincenzopalazzo:
Review fixes pushed in ced9c41 and 2f50922: user secrets use create_new and mode 0600 on Unix, unknown tower API errors retain appointments, outbox entries are fsynced before atomic rename, corrupt committed entries fail closed, and files close before rename. cargo check, six watchtower unit tests, and the first fix-delta review passed. Remaining blockers include crash consistency between monitor persistence and the outbox, startup capture for existing monitors, and skipped breach-path CI.

PR #559, opened 2026-08-12
lampod: reconnect to channel counterparties automatically

vincenzopalazzo:
## Summary

- LDK never redials anyone. After a restart or any TCP drop, a lampo node with live, ready channels sits at **zero peers forever** — it stops forwarding and cannot be paid until an operator manually re-issues `connect`. Observed live: two nodes with a ready channel, both restarted, both at `peers=0` indefinitely.
- `ConnectionNeeded` does not cover this: it fires only when LDK has an onion message to deliver *and* an address to hand us. It is a message-delivery hook, not a reconnect loop.
- Adds lampo's equivalent of ldk-node's `PEER_RECONNECTION_INTERVAL` task: every 10s, dial every disconnected channel counterparty at the last-known address in `peers.json`.
- Bound each attempt at 5s so one hung TCP cannot stall the rest of the channel list. A close before handshake is a failed connect, not success.
- Announced peers are already handled by `ConnectionNeeded`. The store is what matters for unannounced peers — the common case, and the observed incident.

## Reproduction

Two daemons on signet, one ready channel between them, both restarted:

```
peers before: swapd=0 r=0     # indefinitely, channel ready=True the whole time
```

A BOLT12 `fetchinvoice` then fails with "timeout while waiting for the invoice, the offer issuer may be offline" — with both nodes up and the channel fine. Manual `connect` fixes it instantly, which is exactly what software should be doing.

## Test

`channel_peer_reconnects_after_disconnect`: open a channel, sever the connection via `disconnect`, assert the node re-establishes it with no RPC and no manual dial. The nodes have no announce address, so the test specifically proves the persisted-store path.

- with the loop: `ok ... finished in 14.45s` (one ~10s tick)
- with the loop removed: `FAILED ... finished in 54.50s` (retries exhausted)

## Decision Log

**Hardest decision:** where addresses come from. ldk-node persists a peer store; the graph alone is tempting (no new file) but useless for unannounced peers — reconnect-via-graph would NOT have fixed the observed incident. So: `peers.json` written on every successful outbound connect. Graph walk dropped: `ConnectionNeeded` already dials announced TCP addrs.

**Alternatives rejected:**
- Reconnect on `PeerDisconnected` events instead of polling: no such reliable hook covers process restart, which is the main case.
- Persisting inbound peers' socket addresses: an inbound remote port is ephemeral, not their listening port — storing it would produce confident dials to a dead port.
- Graph as fallback: extra code that cannot reach the peers this feature exists for.

**Least confident about:** serial dials inside one tick. A stale stored address costs a TCP timeout per peer; the 5s bound keeps that from stalling the rest of the list.

## Test plan

- [ ] CI passes
- [x] New test red without the loop, green with it (verified both ways)
- [x] `cargo check` clean

### DIFF
--- a/lampod/src/ln/peer_manager.rs
+++ b/lampod/src/ln/peer_manager.rs
@@ -168,6 +168,57 @@ impl LampoPeerManager {
             .unwrap_or_else(|| "127.0.0.1".to_string());
         let bind_addr = format!("{bind_host}:{listen_port}");
 
+        // Reconnect to channel counterparties. LDK never redials anyone:
+        // after a restart (or any TCP drop) a node with live, ready channels
+        // sits at zero peers forever -- it stops forwarding and cannot be
+        // paid -- unless something above it re-establishes the connections.
+        // `ConnectionNeeded` does not cover this: it only fires when LDK has
+        // an onion message to deliver AND knows an address. This loop is
+        // lampo's equivalent of ldk-node's PEER_RECONNECTION_INTERVAL task:
+        // every tick, dial every disconnected channel counterparty at the
+        // address we last reached them on. Announced peers are already
+        // handled by `ConnectionNeeded`.
+        {
+            let peer_manager = peer_manager.clone();
+            let chan_manager = chan_manager.clone();
+            let shutdown = shutdown.clone();
+            let store_path = peer_store_path(&self.conf);
+            tokio::spawn(async move {
+                let mut interval = tokio::time::interval(Duration::from_secs(10));
+                interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+                loop {
+                    interval.tick().await;
+                    if shutdown.load(Ordering::Acquire) {
+                        break;
+                    }
+                    let known = load_peers(&store_path);
+                    for channel in chan_manager.manager().list_channels() {
+                        let peer = channel.counterparty.node_id;
+                        if peer_manager.peer_by_node_id(&peer).is_some() {
+                            continue;
+                        }
+                        let Some(host) = known
+                            .get(&peer.to_string())
+                            .and_then(|addr| addr.parse().ok())
+                        else {
+                            continue;
+                        };
+                        log::info!(
+                            target: "lampo",
+                            "reconnecting to channel peer `{peer}` at `{host}`"
+                        );
+                        // Bound each attempt so one hung TCP cannot stall
+                        // the rest of the channel list.
+                        let _ = tokio::time::timeout(
+                            Duration::from_secs(5),
+                            dial(peer_manager.clone(), peer, host),
+                        )
+                        .await;
+                    }
+                }
+            });
+        }
+
         tokio::spawn(async move {
             log::info!(target: "lampo", "Listening for in-bound connection on {bind_addr}");
             let listener = match tokio::net::TcpListener::bind(bind_addr.clone()).await {
@@ -272,30 +323,17 @@ impl LampoPeerManager {
     }
 
     pub async fn connect(&self, node_id: NodeId, host: SocketAddr) -> error::Result<()> {
-        let Some(close_callback) = net::connect_outbound(self.manager(), node_id, host).await
-        else {
-            log::warn!("impossible connect with the peer `{node_id}`");
-            error::bail!("impossible connect with the peer `{node_id}`");
-        };
-        let mut connection_closed_future = Box::pin(close_callback);
-        let manager = self.manager();
-        loop {
-            match futures::poll!(&mut connection_closed_future) {
-                std::task::Poll::Ready(_) => {
-                    log::info!("node `{node_id}` disconnected");
-                    return Ok(());
-                }
-                std::task::Poll::Pending => {}
-            }
-            // Avoid blocking the tokio context by sleeping a bit
-            match manager.peer_by_node_id(&node_id) {
-                Some(_) => return Ok(()),
-                None => tokio::time::sleep(Duration::from_millis(10)).await,
-            }
-        }
+        dial(self.manager(), node_id, host).await?;
+        // Remember where this peer lives. LDK does not redial anyone on
+        // restart, and a peer that never announced an address -- the common
+        // case for an unannounced node -- is unreachable through the network
+        // graph, so this file is the only durable record of how to get back
+        // to a channel counterparty. The reconnect loop reads it.
+        remember_peer(&peer_store_path(&self.conf), &node_id, &host);
+        Ok(())
     }
 
-    async fn disconnect(&self, node_id: NodeId) -> error::Result<()> {
+    pub async fn disconnect(&self, node_id: NodeId) -> error::Result<()> {
         //check the pubkey matches a valid connected peer
         if self.manager().peer_by_node_id(&node_id).is_none() {
             error::bail!("Error: Could not find peer `{node_id}`");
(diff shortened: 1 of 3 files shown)


lampod/src/ln/peer_manager.rs
```diff
@@ -305,3 +370,57 @@ impl LampoPeerManager {
...
+    let Some(close_callback) = net::connect_outbound(manager.clone(), node_id, host).await else {
+        log::warn!("impossible connect with the peer `{node_id}`");
+        error::bail!("impossible connect with the peer `{node_id}`");
+    };
+    let mut connection_closed_future = Box::pin(close_callback);
+    loop {
+        match futures::poll!(&mut connection_closed_future) {
+            std::task::Poll::Ready(_) => {
+                log::info!("node `{node_id}` disconnected");
+                return Ok(());
```
vincenzopalazzo: **Medium:** `dial` returns `Ok` when TCP dies before handshake.

Then `connect` persists this host into `peers.json`. For an unannounced counterparty that file is the only address we have -- we would redial a dead port forever after restart.

ldk-node treats this path as `Err(ConnectionFailed)`.

```suggestion
            std::task::Poll::Ready(_) => {
                log::info!(target: "lampo", "node `{node_id}` disconnected");
                error::bail!("connection to `{node_id}` closed before handshake");
            }
```

lampod/src/ln/peer_manager.rs
```diff
@@ -168,6 +168,84 @@ impl LampoPeerManager {
...
+                                    _ => {}
+                                }
+                            }
+                        }
+                        for host in candidates {
+                            log::info!(
+                                target: "lampo",
+                                "reconnecting to channel peer `{peer}` at `{host}`"
+                            );
+                            if dial(peer_manager.clone(), peer, host).await.is_ok()
```
vincenzopalazzo: **Medium:** one hung `dial` stalls every other channel peer.

`dial` loops until handshake or the TCP session closes. A host that accepts and then sits there (stale IP, something that never speaks Lightning) blocks this tick *and every later tick* -- we never reach the next counterparty.

Bound it:

```suggestion
                            let _ = tokio::time::timeout(
                                Duration::from_secs(5),
                                dial(peer_manager.clone(), peer, host),
                            )
                            .await;
                            if peer_manager.peer_by_node_id(&peer).is_some()
```

Or spawn and move on. Either way the loop has to stay live.

lampod/src/ln/peer_manager.rs
```diff
@@ -168,6 +168,84 @@ impl LampoPeerManager {
...
+                                    ldk::ln::msgs::SocketAddress::TcpIpV6 { addr, port } => {
+                                        candidates.push(SocketAddr::from((
+                                            std::net::Ipv6Addr::from(*addr),
+                                            *port,
+                                        )))
+                                    }
+                                    _ => {}
+                                }
+                            }
+                        }
```
vincenzopalazzo: **Suggestion:** drop the graph walk.

ldk-node's `PEER_RECONNECTION_INTERVAL` does not consult gossip -- it just walks the peer store. The incident and the new test are both store-only. Graph addresses already get dialled from `ConnectionNeeded` when LDK has a message.

This is also a copy of the IPv4/IPv6 match in `handler.rs`. CLAUDE.md is write-for-today: inbound-only announced peers can be a follow-up.

Store-only loop is ~15 lines:

```rust
let Some(host) = known.get(&peer.to_string()).and_then(|addr| addr.parse().ok()) else {
    continue;
};
log::debug!(target: "lampo", "reconnecting to channel peer `{peer}` at `{host}`");
let _ = tokio::time::timeout(
    Duration::from_secs(5),
    dial(peer_manager.clone(), peer, host),
)
.await;
```

lampod/src/ln/peer_manager.rs
```diff
@@ -168,6 +168,84 @@ impl LampoPeerManager {
...
+                        for host in candidates {
+                            log::info!(
+                                target: "lampo",
+                                "reconnecting to channel peer `{peer}` at `{host}`"
+                            );
```
vincenzopalazzo: **Cleanup:** this fires every 10s for every disconnected peer.

A dead address will print forever. CLAUDE.md: most logs should be debug. Keep `info` for a successful handshake if we want a breadcrumb.

```suggestion
                            log::debug!(
                                target: "lampo",
                                "reconnecting to channel peer `{peer}` at `{host}`"
                            );
```

lampod/src/ln/peer_manager.rs
```diff
@@ -305,3 +370,57 @@ impl LampoPeerManager {
...
+        .and_then(|content| lampo_common::json::from_str(&content).ok())
+        .unwrap_or_default()
+}
+
+fn remember_peer(path: &std::path::Path, node_id: &NodeId, host: &SocketAddr) {
+    let mut peers = load_peers(path);
+    peers.insert(node_id.to_string(), host.to_string());
+    match lampo_common::json::to_string_pretty(&peers) {
+        Ok(json) => {
+            if let Err(err) = std::fs::write(path, json) {
```
vincenzopalazzo, line 394: **Suggestion (non-blocking):** read-modify-write on `peers.json` is racy.

Two concurrent `connect` RPCs can both read the old map and the last writer drops the other peer. Also `write` truncates in place, so a crash mid-write leaves invalid JSON and `load_peers` silently returns empty.

Not today's bug (reconnect loop does not write). If we touch this file again: one `Mutex` around the store + write to `peers.json.tmp` then `rename`.

vincenzopalazzo (commented):
Self-review: I found 2 medium issues in `dial`, and the reconnect path is more code than the bug needs.

Production Readiness: GO

- Medium: `dial` returns `Ok` when TCP dies before the handshake, then `connect` persists that host
- Medium: one hung `dial` stalls every other channel peer for the life of the process

## Code Review

### Positive Aspects
- Diagnosis is right. `ConnectionNeeded` is a message-delivery hook, not a reconnect loop — it cannot fix a restart with unannounced channel peers. Persist last-known address is the only thing that would have unblocked the incident.
- Test is the right one: unannounced nodes, no RPC, just `disconnect` and wait. Red without the loop, green with it.
- Pulling the handshake wait into `dial` so RPC `connect` and the loop share one path is correct.

---

### Issues & Suggestions

#### 1. `dial` reports success when the connection dies before handshake (Medium)

`peer_manager.rs:388-390` returns `Ok(())` if `connection_closed_future` fires before `peer_by_node_id` ever sees the peer. Then `connect` (`:353-359`) treats that as success and writes the host into `peers.json`.

Codex already flagged this and I want to second it — blast radius is worse than a bad RPC return. For an unannounced counterparty the store is the *only* address we have. Persist a host that accepted TCP and then hung up, and after restart we confidently redial a dead port forever.

ldk-node does the opposite here: connection-closed-before-handshake is `Err(ConnectionFailed)`.

**Suggestion:**
```rust
std::task::Poll::Ready(_) => {
    log::info!(target: "lampo", "node `{node_id}` disconnected");
    error::bail!("connection to `{node_id}` closed before handshake");
}
```

#### 2. One hung `dial` stalls the whole reconnect loop (Medium)

`peer_manager.rs:238` awaits each candidate serially, with no timeout. `dial` loops until handshake *or* the TCP session closes. A host that accepts and then sits there (stale IP, greeter that never speaks Lightning) blocks this tick *and every later tick* — we never reach the next channel counterparty.

Same Codex P1, seconding it. Simplest fix is a timeout around the attempt, not a connection manager:

```rust
let _ = tokio::time::timeout(Duration::from_secs(5), dial(peer_manager.clone(), peer, host)).await;
```

Or spawn the dial and move on. Either way the loop has to stay live.

#### 3. Graph fallback is more than this PR needs (Suggestion)

`peer_manager.rs:207-232` walks `announcement_info` and hand-rolls IPv4/IPv6. ldk-node does **not** consult the graph in `PEER_RECONNECTION_INTERVAL` — it just walks the peer store. The incident and the new test are both store-only. Graph addresses already get dialled from `ConnectionNeeded` when LDK has a message.

CLAUDE.md is "write for today". Inbound-only announced peers can be a follow-up. Drop the walk and the loop is ~15 lines:

```rust
let known = load_peers(&store_path);
for channel in chan_manager.manager().list_channels() {
    let peer = channel.counterparty.node_id;
    if peer_manager.peer_by_node_id(&peer).is_some() {
        continue;
    }
    let Some(host) = known.get(&peer.to_string()).and_then(|addr| addr.parse().ok()) else {
        continue;
    };
    log::debug!(target: "lampo", "reconnecting to channel peer `{peer}` at `{host}`");
    let _ = tokio::time::timeout(
        Duration::from_secs(5),
        dial(peer_manager.clone(), peer, host),
    )
    .await;
}
```

That is the whole feature. Also dedup by `node_id` if we ever have two channels with the same peer — today we would dial twice.

#### 4. Reconnect attempts log at `info` every 10s (Cleanup)

`peer_manager.rs:234-237`. A peer whose address is dead will print this forever. CLAUDE.md: most logs should be debug. Keep `info` for a successful handshake if we want a breadcrumb.

#### 5. Squash `chore: cargo fmt` (Cleanup)

CLAUDE.md: never leave a formatting fixup as its own commit. Fold it into `fb43f711`. Subject is also 57 chars — `lampod: auto-reconnect channel peers` fits the 50-char rule.

---

### Summary

Concept ACK. The persist+poll shape is right. Before merge: make `dial` fail on pre-handshake close, bound each attempt, drop the graph walk. Then this is the simple loop it should have been.

---

### Decision Log

**Hardest decision:** where addresses come from. Graph-only is tempting (no new file) but would **not** have fixed the incident — unannounced peers have nothing in the graph. So `peers.json` on successful outbound `connect` is the real feature. I then added the graph walk "for inbound peers that later announced" and that is the overdesign. Should have shipped store-only.

**Alternatives rejected:**
- Reconnect on `PeerDisconnected`: does not cover process restart, which is the main case.
- Persist inbound socket addresses: remote port is ephemeral, not their listen port — we would dial dead ports with confidence.
- Event-driven only via `ConnectionNeeded`: already in tree, and it is a message-delivery hook. Does not run after a quiet restart.

**Least confident about:** serial dials with no timeout. A large channel set plus one black-hole address makes ticks arbitrarily slow. ldk-node has the same serial shape, but they at least fail the attempt when the TCP session dies. I would re-test with a host that accepts and never handshakes before calling this done.

PR #560, opened 2026-08-12
swapd: expire unpaid hold swaps and release their payment hash

vincenzopalazzo:
## Summary

`HoldInvoiceIssued` had no expiry arm. The comment said to let the invoice lapse on the node — but the **swap record itself never ended**. It stayed in `pending()` forever, and because `create_hold_swap` refuses to reuse a payment hash, every abandoned quote **burned its hash permanently** and added a record to every future reconcile pass.

Fail the swap once the node's own hold window has lapsed. The deadline is the window the invoice was issued with (`spark_htlc_expiry_secs + LN_HOLD_MARGIN_SECS`), past which the node will not accept a payment against that hash, so the record is dead by construction.

> Based on `feat/bolt12-swap-primitives` (#553), not `main` — the `lampo-spark-swapd` crate only exists on that branch. This stacks on #553 rather than disturbing it.

## The three things this has to get right

It touches a hash a counterparty may still pay, so the ordering is the whole design:

1. **A payment can land between the deadline check and the release.** Ask the node what it is actually holding: if the hash is `Held`, this is a live swap with their money — it goes to `LnHeld` and we deliver, never expire.
2. **Release the hash with `hold_fail` BEFORE forgetting the record.** A payment arriving against a hash we have already dropped would be held with nobody left to settle or fail it.
3. **Only transition once the node confirms the hold is gone.** If the release fails, stay pending and retry. Keeping a record costs a reconcile pass; abandoning a hash the node still holds can strand a counterparty's payment. Staying is the safe direction.

## Tests

`hold_invoice_action` is a pure decider beside the existing `quoted_action`, unit tested at the boundary, past it, at a zero window, and — the one that matters — with a `created_at` in the future. A backwards clock jump must **wait**, never expire early, because expiring is the irreversible arm; `saturating_add` is what guarantees that. (The test caught me here: my first assertion said such a jump should expire. The code was right and the assertion was wrong.)

The e2e test stages exactly what a restart would see: issue a real hold invoice on a real node, age the record past its window on disk, reconcile with a fresh engine. It asserts **both halves** — the swap reaches `Failed` *and* the node no longer holds the hash. Asserting only the state change would pass even if the release silently did nothing, which is the failure mode that actually strands money.

The hold window is an hour by construction, so waiting it out in a test is not an option; staging the record is the same technique `direction_a_recovers_a_crashed_claim` already uses.

## Decision Log

**Hardest decision:** what to do when `hold_fail` fails. Marking the swap `Failed` anyway is tidier and keeps the store small, but it drops our record of a hash the node may still be holding — so a later payer gets held with no one to settle them. Chose to stay pending and retry forever: an unbounded pending record is a nuisance, a stranded payment is a loss.

**Alternatives rejected:**
- Expire on the invoice's own expiry rather than the hold window: the two can differ, and the hold window is the one that governs whether a payment can still arrive.
- Skip the second `list_holds` check and trust `hold_fail`'s return: it returns before the node has necessarily dropped the hold, so trusting it reintroduces the exact race the arm exists to close.

**Least confident about:** swaps whose `payment_hash` is `None` (Direction B records live under the offer id until a payer appears). They are failed without a release, which is correct — there is no hash to hold against — but that path has no e2e coverage, only the reasoning.

## Test plan

- [ ] CI passes
- [x] `hold_invoice_action` unit tests (27 lib tests green)
- [x] `cargo check` and `cargo fmt` clean; no new clippy warnings in touched files
- [x] `unpaid_hold_swap_expires_and_releases_its_hash` on the regtest operator stack: `ok ... finished in 12.91s` — swap reached `Failed` and the node no longer held the hash


vincenzopalazzo:
Folded into #553 as a single commit: `e9a31774` — *swapd: expire unpaid hold swaps and release their payment hash*.

Same change, cherry-picked onto the rebased swap branch so the fix travels with the crate it belongs to rather than as a stacked PR. The e2e (`unpaid_hold_swap_expires_and_releases_its_hash`) came across with it and still passes against the regtest operator stack; the 35 swapd unit tests are green.

Closing in favour of #553.

PR #562, opened 2026-08-14
lampod: bound the pay RPC wait with a timeout

vincenzopalazzo:
## Problem

Found by the regtest soak simulation (round 89, after 88 successful payments): when the terminal `PaymentEvent` is never generated, the `pay` RPC blocks **forever** — the code even carries the FIXME `this will loop when the Payment event is not generated`.

Repro (sim server, SEED=1337): SIGKILL a node with open channels, relaunch, immediately pay an invoice from it. LDK logs:

```
ERROR ldk lightning::chain::chainmonitor: Failed to update channel monitor: no such monitor registered.
WARN  lampod::jsonrpc::offchain Waiting for payment event....
```

…no failure event ever fires, and the HTTP call never responds (curl times out with an empty reply).

## Fix

Wrap the `events.recv()` loop in `tokio::time::timeout` (120 s): the RPC returns a clear JSON-RPC error (`payment `<id>` did not complete within 120s`) while the payment may still be retried in the background.

## Note for follow-up

The trigger exposes a deeper issue worth its own investigation: after an unclean restart LDK reports `no such monitor registered` and the payment stalls without any failure event — the timeout makes lampo degrade gracefully, but the monitor/persistence path should be looked at separately.

Artifacts: `~/lampo-sim/sim-run/artifacts/20260814-103619-*` on the sim server.

### DIFF
--- a/lampod/src/jsonrpc/offchain.rs
+++ b/lampod/src/jsonrpc/offchain.rs
@@ -1,5 +1,6 @@
 //! Offchain RPC methods
 use std::str::FromStr;
+use std::time::Duration;
 
 use lampo_common::event::ln::LightningEvent;
 use lampo_common::event::Event;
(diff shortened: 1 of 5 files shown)


lampod/src/jsonrpc/offchain.rs
```diff
@@ -122,13 +132,30 @@ pub async fn json_pay(ctx: &LampoDaemon, request: &json::Value) -> Result<json::
...
+            .await
+            .map_err(|_| {
+                Error::Rpc(RpcError {
+                    code: -1,
+                    message: format!(
+                        "payment `{}` did not complete within {}s (no Payment event; \
+                     the payment may still be retried in the background)",
+                        payment_id,
+                        PAY_EVENT_TIMEOUT.as_secs()
+                    ),
```
vincenzopalazzo: Mh this is wrong! we should return an error only when the paymet is failing so probably we should wait for some ldk event stuff

lampod/src/jsonrpc/offchain.rs
```diff
@@ -122,13 +132,28 @@ pub async fn json_pay(ctx: &LampoDaemon, request: &json::Value) -> Result<json::
...
+    // PAY_EVENT_TIMEOUT instead of blocking the caller forever.
     loop {
         log::warn!("Waiting for payment event...");
-        let event = events.recv().await.ok_or(Error::Rpc(RpcError {
-            code: -1,
-            message: format!("No event received, communication channel dropped"),
-            data: None,
-        }))?;
+        let event = tokio::time::timeout(PAY_EVENT_TIMEOUT, events.recv())
+            .await
```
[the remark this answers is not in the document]
vincenzopalazzo: Good catch -- `wait_for_payment_result` now sets one deadline with `Instant::now() + timeout` and uses `timeout_at(deadline, events.recv())`, so unrelated bus events no longer reset the timer. **Fixed in 8caff66.**

lampod/src/jsonrpc/offchain.rs
vincenzopalazzo: Agreed -- the RPC error is only when no terminal `PaymentEvent` arrives before the deadline (status unknown, payment may still retry). LDK `PaymentFailed` still returns a normal `PayResult` with `state: Failure`. Callers pick the wait budget with `timeout: fast|medium|large` (30s / 120s / 600s, default medium). **Fixed in 8caff66.**

PR #564, opened 2026-08-14
lampod: register persisted channel monitors on restart

### LINKED ISSUE #563: Channels silently broken after unclean restart: 'no such monitor registered', payments stall with no failure event
## Summary

Found by the regtest soak simulation (deterministic repro, see below). After a node is SIGKILLed with open channels and relaunched, its channels appear healthy (`peers`/`channels` in `getinfo`), but **payments to, from, and through it stall forever without any failure event**. LDK logs at restart-time:

```
ERROR ldk lightning::chain::chainmonitor: Failed to update channel monitor: no such monitor registered.
```

## Deterministic reproduction

On the sim server: `NODES=3 ROUNDS=0 CHAOS_EVERY=8 SEED=1337 ~/lampo-sim-harness/simulate.sh`
- round 88: chaos `restart9` SIGKILLs n2 (channels open) and relaunches it
- round 89 (n2→n3 invoice) and 90 (n2→n3 offer): pay stalls — route is found (`Got route: path 0…`) but no `PaymentEvent` ever fires
- round 92 (n3→n2 offer): receiving also stalls
- before round 88 the same payments succeeded repeatedly (88/88 rounds green)

Artifacts per failure in `~/lampo-sim/sim-run/artifacts/20260814-1122*` (full node dirs incl. `manager`, `network_graph`, logs).

## Related

- #562 bounds the `pay` RPC with a 120s timeout so callers get an error instead of an infinite hang (verified firing: 4 bounded errors in the run above) — this issue tracks the underlying monitor/persistence problem.
- #561 (fundchannel panic) was found by the same soak.

## Suspect area

Channel monitor persistence/restoration on unclean shutdown (`manager` file / chainmonitor registration at startup). After the SIGKILL the monitor list appears inconsistent: the channel manager lists channels whose monitors are not registered, so commitment/HTLC updates fail silently.

vincenzopalazzo:
## Problem

Issue #563: after a node restart, all channels that existed before the restart are *silently broken*:
- payments from, to, or routed through the node stall forever with **no failure event**
- LDK logs `Failed to update channel monitor: no such monitor registered` on the first monitor update
- peers keep reconnecting every ~120s, exchange Init + data_loss_protect, and never make progress

Reproduced deterministically by the regtest soak on the debian server (restart9 chaos = SIGKILL + cold relaunch): SEED=1337 round 88, reconfirmed on SEED=4242 rounds 25-39.

## Root cause

On restart (`LampoChannelManager::restart`) we read the `manager` file and the channel monitors from disk and hand the monitors to `ChannelManagerReadArgs` — but `ChannelManagerReadArgs` only uses them to rebuild the `ChannelManager`. LDK's docs require the caller to *move the monitors into the `chain::Watch` (the `ChainMonitor`)* after restoring:

```rust
// Move the monitors to the ChannelManager's chain::Watch parameter
for monitor in channel_monitors {
    chain_monitor.watch_channel(monitor.channel_id(), monitor);
}
```

We never did, so the `ChainMonitor` starts empty and the first monitor update of every pre-existing channel fails. This hits **clean restarts as much as SIGKILL ones** — the SIGKILL just made it visible sooner — and the unregistered monitors also cannot react to chain events for those channels.

## Fix

Register every persisted monitor with the `ChainMonitor` right after `ChannelManager::read`, logging success/failure per channel.

## Validation

Regression re-run of the same deterministic chaos schedule on the sim cluster after this patch: channels survive a restart, payments from/to/through the restarted node succeed, no `no such monitor registered` in the logs (details in the sim run log).

Fixes #563

### DIFF
--- a/lampod/src/ln/channel_manager.rs
+++ b/lampod/src/ln/channel_manager.rs
@@ -18,7 +18,7 @@ use lampo_common::ldk::block_sync::BlockSource;
 use lampo_common::ldk::chain::chaininterface::{BroadcasterInterface, FeeEstimator};
 use lampo_common::ldk::chain::chainmonitor::ChainMonitor;
 use lampo_common::ldk::chain::channelmonitor::ChannelMonitor;
-use lampo_common::ldk::chain::BlockLocator;
+use lampo_common::ldk::chain::{BlockLocator, Watch};
 use lampo_common::ldk::ln::channelmanager::{ChainParameters, ChannelManagerReadArgs};
 use lampo_common::ldk::onion_message::messenger::DefaultMessageRouter;
 use lampo_common::ldk::routing::gossip::NetworkGraph;
@@ -320,7 +320,6 @@ impl LampoChannelManager {
 
         let _ = self.network_graph();
         let monitors = self.get_channel_monitors()?;
-        let monitors = monitors.iter().collect::<Vec<_>>();
 
         let default_message_router = DefaultMessageRouter::new(
             self.graph(),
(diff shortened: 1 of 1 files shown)


vincenzopalazzo:
Verified on the debian pre-production cluster (branch `sim/main-next` harness):

**Restart regression** (the scenario that failed on current main):
- 3-node regtest cluster (channels ready), SIGKILL all nodes, relaunch on the same data dirs with this PR built (`c82bae0`)
- all nodes back with identical node ids; invoice payment `n1->n2` right after restart: `state: Success` + full preimage in **4.8 s**
- `grep -c "no such monitor registered"` across all node logs: **0 / 0 / 0**

Same scenario on unfixed main: payment hangs until the client times out (120 s, 0 bytes) and `Failed to update channel monitor: no such monitor registered` fires on the first HTLC update (artifacts: `~/lampo-main-sim/sim-run/artifacts/20260815-142859-*` on the server).

Note: #580 is a byte-identical re-land of this patch — consolidating there.

Issue #567, opened 2026-08-15
keysend returns {} and there is no payment-status RPC: keysend outcomes are unauditable


vincenzopalazzo:
## Problem

Payments made via the `keysend` RPC are unauditable:

1. `json_keysend` (`lampod/src/jsonrpc/offchain.rs`) returns `{}` and discards the `PaymentHash` produced by `OffchainManager::keysend` — the caller cannot correlate the payment with anything.
2. There is no `listpayments` / payment-status RPC at all (HTTP surface: channels, close, connect, decode, fundchannel, funds, getinfo, invoice, networkchannels, new_addr, offer, pay). Once `pay`/`keysend` returns, the outcome of a pending/failed payment can no longer be observed over RPC.

Found by the simulation harness: keysend rounds are marked "OK 0s" purely because the RPC returned without an error — even when the destination node's channels were silently broken (issue #563) and the payment could not possibly have settled. The harness cannot detect this.

## Proposal

- `keysend` returns `{ payment_hash, payment_id, state }` (stop discarding the hash; `payment_id` is already known inside `OffchainManager::keysend`)
- Add `listpayments` (and/or `payment_status <hash>`) so wallets/tests can observe `Pending/Succeeded/Failed` including the failure reason, matching the events the node already processes internally

Without this, keysend failures are invisible to any RPC client — including production users.

PR #568, opened 2026-08-15
lampod: return the full payment result from keysend

vincenzopalazzo:
## Problem

Part 1 of issue #567: `keysend` returned `{}` and discarded the `PaymentHash`, so keysend outcomes were unauditable — callers could not correlate the payment or tell whether it settled. The simulation harness marked keysend rounds "OK" purely on RPC acceptance, even while the destination node's channels were silently broken (#563) and the payment could not possibly have settled.

## Change

Mirror `json_pay`'s event handling in `json_keysend`:

- subscribe to the event bus before initiating (avoid the missed-event race)
- match `PaymentReceipt` (preimage + payer proof) and the terminal `PaymentEvent` **for our payment id** — for keysend, `PaymentId` is the payment hash
- answer with the same shape as `pay`: `{ state, path, payment_hash, payment_preimage, payer_proof }`
- bound the wait at 120s (same rationale as #562's pay bound: stall after unclean restart must not block the RPC forever)

**Breaking change**: response shape changes from `{}` to the `PayResult` object.

## Validation

- cargo check clean
- the soak harness now verifies keysend like invoice/offer payments (state + preimage) once this lands — regression rerun recorded in the sim run log after ship

Refs #567 (the payment-status/listpayments RPC remains open there).

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,12 @@ following command to use the example config file:
 cp lampo.example.conf ~/.lampo/signet/lampo.conf
 ```
 
+A newly generated wallet jumps its checkpoint to the chain tip (fast-sync, on
+by default) so it does not scan from genesis. A mnemonic restore never takes
+that automatic jump, even when its local database is new, because the seed may
+own historical funds. Use `reindex=<height>` in `lampo.conf` for a known
+birthday.
+
 Then you can query the node with the following command:
 
 ``` 
--- a/lampo-bdk-wallet/src/lib.rs
+++ b/lampo-bdk-wallet/src/lib.rs
@@ -1,4 +1,6 @@
 //! Wallet Manager implementation with BDK
+use std::fs;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 // The on-chain wallet + its sqlite handle are guarded by a std (blocking)
 // mutex, not a tokio one: their critical sections are short and fully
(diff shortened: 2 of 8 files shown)


vincenzopalazzo:
Heads-up: #585 (same route registration + `PayResult` shape, found independently by the main-next pre-production soak) was just merged, so the remaining value of this PR is the **120s bounded event wait** (`tokio::time::timeout` around `events.recv()`) — which is strictly better than the merged unbounded loop. Suggest rebasing this to just that change (or applying the bound to `wait_for_payment_result` in `lampod/src/jsonrpc/offchain.rs`, now shared by pay+keysend).

Verification for the record: built 589b824 on the debian regtest cluster, `SEED=42 NODES=3 ROUNDS=2` → round-2 keysend `n3->n2` `Success` + preimage in 10s (fails empty on pre-#585 main).

lampod/src/jsonrpc/offchain.rs
```diff
@@ -163,10 +164,77 @@ pub async fn json_pay(ctx: &LampoDaemon, request: &json::Value) -> Result<json::
...
+        let event = tokio::time::timeout(KEYSEND_EVENT_TIMEOUT, events.recv())
+            .await
+            .map_err(|_| {
+                Error::Rpc(RpcError {
+                    code: -1,
+                    message: format!(
+                        "keysend `{}` did not complete within {}s (no Payment event; \
+                         the payment may still be retried in the background)",
+                        payment_id,
+                        KEYSEND_EVENT_TIMEOUT.as_secs()
```
vincenzopalazzo: This is a bad error to send :/ unfortuantly! but probably it is neeed?

PR #571, opened 2026-08-15
lampod: add a listpayments RPC backed by an in-memory payment history

vincenzopalazzo:
## Problem

Refs #567 (the remaining half): once `pay`/`keysend` returned, payment outcomes were no longer observable anywhere over RPC — no way to list payments, no way to audit what a node sent/received. The LDK `PaymentClaimed` handler carried the FIXME "make persistent these information".

## Change

- New `PaymentReceived` event emitted from the `PaymentClaimed` handler (hash, amount, preimage when available) so inbound payments reach the event bus
- A subscriber task (`lampod::payments::record_payments`) keeps a **capped (1000), in-memory** history:
  - **sent**: terminal `PaymentEvent` (state, hash, path, failure reason) with the preimage attached from the preceding `PaymentReceipt`
  - **received**: `PaymentReceived` (hash, amount, preimage)
- New `listpayments` RPC (JSON-RPC + HTTP route)

Boot-scoped by design — an audit/debug surface, not an accounting source of truth (persistence intentionally out of scope). Sent-payment events don't carry the amount, so only received records have `amount_msat`.

## Validation

After ship: `listpayments` on a soaking node returns the expected records with real preimages, in both directions (regression logged in the sim run log).

### DIFF
--- a/lampod/src/jsonrpc/offchain.rs
+++ b/lampod/src/jsonrpc/offchain.rs
@@ -186,3 +186,13 @@ pub async fn json_keysend(ctx: &LampoDaemon, request: &json::Value) -> Result<js
     let payment_id = hex::encode(payment_id.0);
     wait_for_payment_result(events, &payment_id).await
 }
+
+/// List the in-memory payment history (issue #567): every terminal sent
+/// payment and every claimed inbound payment since node start.
+pub async fn json_listpayments(
+    ctx: &LampoDaemon,
+    _request: &json::Value,
+) -> Result<json::Value, Error> {
+    log::info!("call for `listpayments`");
+    Ok(json::to_value(ctx.payments().list())?)
+}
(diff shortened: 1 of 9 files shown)


lampod/src/jsonrpc/offchain.rs
```diff
@@ -170,3 +170,13 @@ pub async fn json_keysend(ctx: &LampoDaemon, request: &json::Value) -> Result<js
...
 }
+
+/// List the in-memory payment history (issue #567): every terminal sent
+/// payment and every claimed inbound payment since node start.
+pub async fn json_listpayments(
+    ctx: &LampoDaemon,
+    _request: &json::Value,
+) -> Result<json::Value, Error> {
+    log::info!("call for `listpayments`");
+    Ok(json::to_value(ctx.payments().list())?)
```
vincenzopalazzo, line 197: We should add pagination to the payments

lampod/src/payments.rs
vincenzopalazzo, line 1: it is possible to do this with a minimal persistance?

vincenzopalazzo (commented):
self review landed

vincenzopalazzo:
Waiting for this https://github.com/vincenzopalazzo/lampo.rs/pull/584 and including this functionality in that PR

PR #573, opened 2026-08-15
wallet: never let a funding tx locktime run ahead of the chain tip (fixes #572)

### LINKED ISSUE #572: Funding tx broadcast rejected as non-final: locktime taken from an LDK height ahead of the chain tip; channel sits zombie-pending with no error
Found by the new multi-hop sim harness (`sim/multihop.sh`, first run 2026-08-15, regtest, binary 29473fe).

## What happened

1. `fundchannel` hs->hm: `create_transaction` sets the funding tx **nLockTime = LDK `current_best_block().height`** (`lampod/src/actions/handler.rs` ~282 -> `lampo-bdk-wallet/src/lib.rs` `create_transaction`, "discourage fee sniping").
2. LDK broadcast -> bitcoind rejects: **`RPC error -26: non-final`** (`lampo-chain/src/lib.rs:210`). The tx had locktime 14777 while bitcoind's tip was still below that - LDK's height view was **ahead** of the chain tip.
3. The error is logged but the flow continues: `channel pending` - the channel sits **zombie-pending** (no retry from the funding path, no failure event, no error surfaced to the caller of `fundchannel`).
4. Recovery only happened because *another* process kept mining: LDK's channel monitor rebroadcast on a later block-connected event, the tip moved past the locktime, and the funding confirmed ~20 blocks later. On a quiet chain this would hang indefinitely.

Log evidence (hs):

```
06:52:33 INFO lampod::actions::handler propagate funding transaction ...
06:52:33 INFO lampod::actions::handler funding transaction created `4b1776ee...`
06:52:35 INFO ldk Broadcasting funding transaction with txid 4b1776ee...
06:52:35 INFO lampo_chain Broadcasting tx result: Err(Rpc(RpcError { code: -26, message: "non-final" }))
06:52:35 ERROR httpd error from backend code: -1, message: Failed to broadcast transaction: RPC error -26: non-final.
06:52:35 INFO lampod::actions::handler channel pending ... funding `4b1776ee...:0`
```

Decoded tx: `locktime 14777`. The funding eventually confirmed in block 14842 after unrelated mining moved the tip; the channel then became ready by itself (so confirmation processing is fine once the tx actually lands).

This is very likely the same root cause as the 2026-08-14 22:03 watch-item ("funding tx apparently never confirmed - possible broadcast gap"), which was closed as environmental.

## Proposed fix (two layers)

1. `handler.rs`: derive the funding locktime from `min(ldk_height, backend.get_best_block().height)` and subtract 1 - the anti-sniping property is preserved and the tx can never be non-final due to height-view skew.
2. `lampo-bdk-wallet create_transaction`: clamp defensively (`best_block - 1`) with a comment referencing this incident.

Follow-up (separate issue): `brodcast_tx` failure during funding should surface as a channel failure / retriable state instead of a silent zombie pending channel.

Artifacts: `~/lampo-sim/sim-run/artifacts/20260815-025925-mh_cluster_not_established` on the debian sim server.

vincenzopalazzo:
Fixes #572 (full evidence and repro story in the issue).

## Problem

`create_transaction` set the funding tx `nLockTime` to LDK's
`current_best_block().height`. LDK's height view can run **ahead** of the
backend tip while the two catch up at different rates, so `sendrawtransaction`
rejected the funding broadcast with `-26 non-final`. The channel then sat
**zombie-pending**: no retry from the funding path, no failure event, nothing
surfaced to the `fundchannel` caller. It only recovered because unrelated
mining moved the tip past the locktime and LDK's monitor rebroadcast later.

Caught live on the first run of the new `sim/multihop.sh` harness (regtest,
2026-08-15): funding tx `4b1776ee…` locktime 14777 vs a lower chain tip,
rejected, confirmed only ~20 blocks later.

## Fix (two layers)

- `lampod/src/actions/handler.rs`: the funding locktime height is now
  `min(ldk_height, backend_tip) - 1`. The anti-fee-sniping property is
  preserved; the tx can no longer be non-final due to height-view skew.
- `lampo-bdk-wallet/src/lib.rs`: `create_transaction` keeps one block of
  headroom as defense in depth for any other caller.

## Testing

- `cargo check -p lampod -p lampo-bdk-wallet` clean.
- Regression on the sim server (same seed) once merged: `sim/multihop.sh`
  build phase no longer produces a zombie-pending channel.

### DIFF
--- a/lampo-bdk-wallet/src/lib.rs
+++ b/lampo-bdk-wallet/src/lib.rs
@@ -282,8 +282,14 @@ impl WalletManager for BDKWalletManager {
         let mut wallet = self.wallet.lock().unwrap();
 
         // We set nLockTime to the current height to discourage fee sniping.
-        let locktime =
-            LockTime::from_height(best_block.to_consensus_u32()).unwrap_or(LockTime::ZERO);
+        // The caller clamps the height against the chain tip (issue #572),
+        // but keep one block of headroom here too: a locktime equal to the
+        // tip can be rejected as `non-final` if the views skew again, and
+        // one block older costs nothing for anti-sniping.
+        let locktime = LockTime::from_height(
+            best_block.to_consensus_u32().saturating_sub(1),
+        )
+        .unwrap_or(LockTime::ZERO);
 
         let mut tx = wallet.build_tx();
         tx.add_recipient(script, amount)
--- a/lampo-bdk-wallet/src/lib.rs
+++ b/lampo-bdk-wallet/src/lib.rs
@@ -286,10 +286,8 @@ impl WalletManager for BDKWalletManager {
         // but keep one block of headroom here too: a locktime equal to the
         // tip can be rejected as `non-final` if the views skew again, and
         // one block older costs nothing for anti-sniping.
-        let locktime = LockTime::from_height(
-            best_block.to_consensus_u32().saturating_sub(1),
-        )
-        .unwrap_or(LockTime::ZERO);
+        let locktime = LockTime::from_height(best_block.to_consensus_u32().saturating_sub(1))
+            .unwrap_or(LockTime::ZERO);
 
         let mut tx = wallet.build_tx();
         tx.add_recipient(script, amount)
--- a/lampo-chain/src/lib.rs
+++ b/lampo-chain/src/lib.rs
@@ -101,6 +101,15 @@ pub struct LampoChainSync {
 }
 
 impl LampoChainSync {
+    async fn broadcast_once(
+        &self,
+        tx: &lampo_common::bitcoin::Transaction,
+    ) -> Result<json::Value, lightning_block_sync::rpc::RpcClientError> {
+        self.rpc_client
+            .call_method::<json::Value>("sendrawtransaction", &[serialize_hex(tx).into()])
+            .await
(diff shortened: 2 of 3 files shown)


vincenzopalazzo:
We need to define if this is possible to happen on mainet

vincenzopalazzo:
## Review vs main + ldk-node

**The bug is real**, not AI-invented. Live evidence in #572 (multihop sim 2026-08-15): funding tx locktime 14777 vs a lower bitcoind tip, `-26 non-final`, then zombie-pending until unrelated mining + monitor rebroadcast.

Still on current `main`:
- handler passes LDK `current_best_block().height` straight into `create_transaction`
- wallet sets `nLockTime` to that height
- `brodcast_tx` does not retry `-26 non-final`

**ldk-node** (`src/event.rs` FundingGenerationReady): same anti-snipe pattern — `LockTime::from_height(channel_manager.current_best_block().height)` with **no clamp**. It heals the same race via later monitor rebroadcast, not by clamping. Lampo's extra `min(ldk, backend)-1` + bounded 3x3s retry is the right production hardening on top of that model.

Rebased onto current `main`. Unit + `fund_a_simple_channel` + `pay_invoice` green.

PR #606, opened 2026-09-12
wallet: report real UTXO confirmation counts

vincenzopalazzo:
## Summary
- `list_transactions` / `funds` hardcoded `confirmed: 0` for every UTXO.
- Derive confirmation count from BDK `chain_position` vs wallet tip so restore and production safety checks can trust `funds`.

## Test plan
- [ ] `cargo check -p lampo-bdk-wallet`
- [ ] On debian mutinynet: `funds` for m1 shows non-zero `confirmed` for the known deep-confirmed UTXO
- [ ] Mempool-only UTXOs still report `confirmed: 0`

Made with [Cursor](https://cursor.com)


vincenzopalazzo:
## Review vs main + ldk-node

**The bug is real.** `list_transactions` / `funds` still hardcodes `confirmed: 0` for every UTXO on `main`. Restore and production-safety checks cannot tell a mempool coin from a deep-confirmed one.

**ldk-node** does not expose per-UTXO confirmation counts. It reports aggregate on-chain totals (`list_balances`) and uses `chain_position.is_unconfirmed()` internally. Deriving count from BDK `confirmation_height_upper_bound` vs wallet tip is the right lampo-shaped fix.

Rebased onto current `main`. `lampo-bdk-wallet` unit + `sweep_funds_after_channel_close` green.

